PR #179, opened 2015-03-03
adapt to new rubocop

jreidinger:
disable one of new style checks for Struct.new

### DIFF
--- a/test/addon_spec.rb
+++ b/test/addon_spec.rb
@@ -169,10 +169,12 @@ describe Registration::Addon do
 
   describe "#selectable?" do
     let(:addons) do
-      Registration::Addon.find_all(double(
-                                     get_addon_list:     [addon_with_child_generator],
-                                     activated_products: []
-        ))
+      Registration::Addon.find_all(
+        double(
+          get_addon_list:     [addon_with_child_generator],
+          activated_products: []
+        )
+      )
     end
 
     let(:parent) { addons.first }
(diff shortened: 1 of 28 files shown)


test/addon_spec.rb
```diff
@@ -171,8 +170,8 @@
   describe "#selectable?" do
     let(:addons) do
       Registration::Addon.find_all(double(
-          get_addon_list:     [addon_with_child_generator],
-          activated_products: []
+                                     get_addon_list:     [addon_with_child_generator],
+                                     activated_products: []
```
mchf: I'd expect:
```
find_all(
  double(
    get_...
    activated_...
  )
)
```
and definitely not so big indentation of params

test/helpers_spec.rb
```diff
@@ -191,10 +191,10 @@
...
 
       addon = Registration::Addon.new(addon_generator(
-          "zypper_name"  => "sle-sdk",
-          "version"      => "12",
-          "arch"         => "x86_64",
-          "release_type" => nil
+                                        "zypper_name"  => "sle-sdk",
+                                        "version"      => "12",
+                                        "arch"         => "x86_64",
+                                        "release_type" => nil
```
mchf: same as above

test/addon_spec.rb
jreidinger: yep, it is indended to align method call, so your one with double is nicer. This is just auto-correct.

jreidinger:
@mchf updated

mchf:
LGTM

jreidinger:
thanks

PR #180, opened 2015-04-07
Register addons from media (FATE#318505)

lslezak:
This is the first part of the complete implementation, I did some additional refactoring and added some more RSpec tests but the complete change would be too hard to review.

This part contains some refactoring anyway (moving the code from the client to a separate class to allow code reusing).

See https://trello.com/c/GGlxEy3u and https://fate.suse.com/318505 for details

### DIFF
--- a/src/lib/registration/registration_ui.rb
+++ b/src/lib/registration/registration_ui.rb
@@ -182,9 +182,11 @@ module Registration
       SwMgmt.set_repos_state(update_repos, false)
     end
 
-    # Register the selected addons, asks for reg. codes if required
-    # @param selected_addons
-    # @param known_reg_codes [Hash]
+    # Register the selected addons, asks for reg. codes if required, known_reg_codes
+    # @param selected_addons [Array<Addon>] list of addons selected for registration,
+    #   successfully registered addons are removed from the list
+    # @param known_reg_codes [Hash] remembered reg. code, it's updated with the
+    #   user entered values
     # @return [Symbol]
     def register_addons(selected_addons, known_reg_codes)
       # if registering only add-ons which do not need a reg. code (like SDK)
--- a/src/lib/registration/registration_ui.rb
+++ b/src/lib/registration/registration_ui.rb
@@ -25,12 +25,14 @@ require "registration/connect_helpers"
 require "registration/registration"
 require "registration/sw_mgmt"
 require "registration/storage"
+require "registration/ui/addon_reg_codes_dialog"
 
 module Registration
   # Registration functions with errror handling, progress messages, etc...
-  # This is a high level APi above Registration::Registration class
+  # This is a high level API above Registration::Registration class
   class RegistrationUI
     include Yast::Logger
+    include Yast::UIShortcuts
     include Yast::I18n
     extend Yast::I18n
 
@@ -41,7 +43,9 @@ module Registration
       textdomain "registration"
       @registration = registration
 
+      Yast.import "Mode"
       Yast.import "Popup"
+      Yast.import "Wizard"
     end
 
     # register the system and the base product
@@ -99,7 +103,7 @@ module Registration
         log.info "System update failed, removing the credentials to register from scratch"
         Helpers.reset_registration_status
         UrlHelpers.reset_registration_url
-        ::Registration::Storage::Cache.instance.upgrade_failed = true
+        Storage::Cache.instance.upgrade_failed = true
       end
 
       updated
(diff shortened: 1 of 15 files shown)


src/lib/registration/registration_ui.rb
```diff
@@ -178,8 +182,100 @@ def disable_update_repos(product_service)
       SwMgmt.set_repos_state(update_repos, false)
     end
 
+    # Register the selected addons, asks for reg. codes if required
+    # @param selected_addons
```
mvidner: [Array\<Addon>]

mvidner:
The code looks well formed, but I have no idea whether it makes sense.

That could be only expected, as I don't know the APIs. But I've realized that it would be much easier for me to learn the problem domain and the APIs if we had yardoc comments, especially short summaries about classes and hyperlinks to related classes.

For example:

`RegistrationUI#register_addons` has an argument `selected_addons`. In the docs for class Addon I could not find the `free` method. (That is probably a generic problem of teaching Yard to understand def_delegators) The class should have a description including a link to the "underlying suse connect object". (I found https://github.com/SUSE/connect but could not add it to http://www.rubydoc.info/github for some reason)

test/sw_mgmt_spec.rb
```diff
@@ -46,6 +46,24 @@
     }
   end
 
+  describe ".init" do
+    it "initializes package management and returns true" do
```
ancorgs: This test basically mimics the implementation line by line. There is no much value in such a test.

test/sw_mgmt_spec.rb
```diff
@@ -46,6 +46,24 @@
...
+      expect(Yast::PackageCallbacks).to receive(:InitPackageCallbacks)
+      expect(Yast::Pkg).to receive(:TargetInitialize)
+      expect(Yast::Pkg).to receive(:TargetLoad)
+      expect(Yast::Pkg).to receive(:SourceRestore).and_return(true)
+
+      expect(Registration::SwMgmt.init).to eq(true)
+    end
+
+    it "returns false when cannot obtain libzypp lock" do
+      expect(Yast::PackageLock).to receive(:Connect).with(false).and_return("connected" => false)
```
ancorgs: I would expect this first ```expect``` to be an ```allow```. :-)

Another thing I would probably do, although is open to discussion because it can be considered overkill and too nested:

```ruby
describe ".init" do
  before do
    allow(Yast::PackageLock).to receive(:Connect).and_return("connected" => connected)
  end

  context "when the libzypp lock can be obtained" do
    let(:connected) { true }

    it "returns the result of Pkg::SourceRestore" do
      # Mock dangerous calls
      allow(Yast::PackageCallbacks).to receive(:InitPackageCallbacks)
      allow(Yast::Pkg).to receive(:TargetInitialize)
      allow(Yast::Pkg).to receive(:TargetLoad)

      expect(Yast::Pkg).to receive(:SourceRestore).and_return("something")
      expect(Registration::SwMgmt.init).to eq "something"
    end
  end

  context "when the libzypp lock cannot be obtained" do
    let(:connected) { false }

    it "returns false" do
       expect(Registration::SwMgmt.init).to eq false
    end
  end
end
```

For me, it makes more clear what you are testing and it's not just a copy&paste of the method implementation with an "expect" for each line.

test/sw_mgmt_spec.rb
```diff
@@ -200,4 +218,40 @@
       Registration::SwMgmt.select_addon_products
     end
   end
+
+  describe ".products_from_repo" do
+    before do
+      expect(Yast::Pkg).to receive(:ResolvableProperties)
```
ancorgs: "expect" inside a "before" is an smell to me.

test/sw_mgmt_spec.rb
```diff
@@ -46,6 +46,24 @@
     }
   end
 
+  describe ".init" do
```
ancorgs: In general this whole file would benefit from using ```subject```

test/sw_mgmt_spec.rb
```diff
@@ -200,4 +218,40 @@
...
+
+  describe ".select_product_addons" do
+    # just the sle-module-legacy product
+    let(:products) { [YAML.load_file(fixtures_file("products_legacy_installation.yml")).first] }
+
+    it "selects remote addons matching the product resolvables" do
+      available_addons = YAML.load_file(fixtures_file("available_addons.yml"))
+
+      # expect the sle-module-legacy product to be selected
+      expect(available_addons[4]).to receive(:selected)
```
ancorgs: NP: this is extremely nitpicking, but it's somehow also related with the idea of testing behavior instead of implementation outlined above.

What about having?
```ruby
  Registration::SwMgmt.select_product_addons(products, available_addons)
  expect(available_addons[4].selected?).to eq true
```

In addition, it will allow you to define the whole thing using contexts to be more declarative... again open to discussion, @jreidinger would not like it :-)

```ruby
describe ".select_product_addons" do
  # just the sle-module-legacy product
  let(:products) { [YAML.load_file(fixtures_file("products_legacy_installation.yml")).first] }
  before do
    Registration::SwMgmt.select_product_addons(products, available_addons)
  end

  context "when there is a remote addon matching the product resolvables" do
    let(:available_addons) { YAML.load_file(fixtures_file("available_addons.yml")) }

    it "selects the addon" do
      expect(available_addons[4].selected?).to eq true
    end

    it "does not select other addons" do
      expect(available_addons[0].selected?).to eq false
    end
  end

  context "when there is no matching remote addon" do
    let(:available_addons) { [] }

    it "reports an error" do
      expect(Yast::Report).to receive(:Error).with(/Cannot find remote product/)
    end
  end
end

File: test/sw_mgmt_spec.rb [hunk 7]
```diff
@@ -200,4 +218,40 @@
...
+      Registration::SwMgmt.select_product_addons(products, available_addons)
+    end
+
+    it "reports an error when the matching remote addon is not found" do
+      available_addons = []
+
+      # expect the sle-module-legacy product to be selected
```
ancorgs: This comment should not be here

ancorgs:
I completely agree with  @mvidner's comment about the lack of yardoc documentation offering an overview on what every class is intended to do and how they work together. I'm a total newcomer to this module and the review is taking me quite some time.

test/media_addon_workflow_spec.rb
```diff
@@ -0,0 +1,49 @@
+#! /usr/bin/env rspec
+
+require_relative "spec_helper"
+require "yaml"
+
+describe "Registration::UI::MediaAddonWorkflow" do
```
ancorgs: I see the same problem here already mentioned in other tests. It actually looks to me just like a 1:1 reproduction of the implementation. But in this case I'm not so certain about how to fix it.

test/sw_mgmt_spec.rb
lslezak: Yes, the function itself is just a simple sequence of package management calls, no rocket science there...

The value is that the function is covered by tests (higher code coverage, that was actually my intent here) and we can at least ensure that there is no syntax error.

Or any idea how to improve it?

test/sw_mgmt_spec.rb
lslezak: Oh, I didn't know that it's possible to use inside a `before` block a value which is defined later inside `context` block, interesting... :astonished:

test/sw_mgmt_spec.rb
lslezak: @ancorgs Um, and what's wrong with it? What should I use instead?

test/sw_mgmt_spec.rb
lslezak: Um, the tricky part is that the `selected` method adds the Addon to the global list (see https://github.com/yast/yast-registration/blob/master/src/lib/registration/addon.rb#L99) 

The reason for 
```ruby
expect(available_addons[4]).to receive(:selected)
```
is to stub the adding (modifying the global variable).  Without it I'd need to add an `after` block to ensure the list is cleared after the test to avoid possibly side effects for the other tests.

test/media_addon_workflow_spec.rb
lslezak: Yes, it's mainly an UI stuff, I'm not sure how to test it better.

test/sw_mgmt_spec.rb
ancorgs: Good point. I forgot that that the actions on ```Addon``` are actually quite persistent.

lslezak:
BTW I'll improve the yardoc comments in a separate PR so we could move forward here...

ancorgs:
Yes, let move forward (but don't discard another round of nitpicking comments about the tests and the ```expect``` abuse) :wink: 

LGTM

lslezak:
@ancorgs thanks for the review!

PR #182, opened 2015-04-09
Collect product rename also for the base product (bsc#925700)

lslezak:
- the internal SLES-for-SAP product identifier has been changed

- 3.1.129.1

See the details in https://bugzilla.novell.com/show_bug.cgi?id=925700

### DIFF
--- a/test/registration_spec.rb
+++ b/test/registration_spec.rb
@@ -36,10 +36,12 @@ describe "Registration::Registration" do
 
     it "adds the selected product and returns added zypp services" do
       product = {
-        "arch"         => "x86_64",
-        "name"         => "sle-sdk",
-        "version"      => "12",
-        "release_type" => nil
+        "arch"              => "x86_64",
+        "name"              => "sle-sdk",
+        "version"           => "12",
+        "release_type"      => nil,
+        "identifier"        => "SLES_SAP",
+        "former_identifier" => "SUSE_SLES_SAP"
       }
 
       service_data = {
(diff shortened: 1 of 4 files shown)


lslezak:
JFYI: The Travis failure is being solved in https://github.com/yast/yast-devtools/pull/83

test/registration_spec.rb
```diff
@@ -58,6 +60,10 @@
       expect(available_addons.find { |addon| addon.identifier == "sle-sdk" }).to \
         receive(:registered)
 
+      # the received product renames are passed to the software management
+      expect(Registration::SwMgmt).to receive(:update_product_renames)
+        .with("SUSE_SLES_SAP" => "SLES_SAP")
+
       allow(File).to receive(:exist?).with(
```
ancorgs: NTH: This is something I'll also comment in the other pull request I'm reviewing right now. I would have expected a different ```it``` to test this.

In this module in general we abuse of huge ```it```s with tons of ```expect``` instead of smaller and more specific ``it```s with only a few ```expect```, using ```allow``` when we need to mock something.

Since this PR is more urgent, I will LGTMed even with this concern. But I will be more strict in the other PRs so we fix it for the future. :-)

ancorgs:
lgtm, for the time being :-)

lslezak:
Yes, I was also thinking about a separate test case, but because this is an urgent fix I preferred a smaller change as possible.

Thanks for the review!

PR #183, opened 2015-04-10
Refactoring

lslezak:
Code refactoring:

- Split complex methods
- Move registration update to a separate class
- Added more tests
- Added more `yardoc` comments

According to CodeClimate the code quality has been [improved by +0.39 to 2.82 GPA](https://codeclimate.com/github/yast/yast-registration/compare/refactoring#ratings).

Note: Adding more `yardoc` comments is still pending, let's review this part...

### DIFF
--- a/.pullreview.yml
+++ b/.pullreview.yml
@@ -1,0 +1,4 @@
+---
+rules:
+  ignore:
+    - indent_when_as_deep_as_case
--- a/src/clients/inst_scc.rb
+++ b/src/clients/inst_scc.rb
@@ -181,25 +181,6 @@ module Yast
       end
     end
 
-    # display the main registration dialog
-    def show_scc_credentials_dialog
-      Wizard.SetContents(
-        # dialog title
-        _("Registration"),
-        scc_credentials_dialog,
-        scc_help_text,
-        GetInstArgs.enable_back,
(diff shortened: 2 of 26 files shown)


lslezak:
[PullReview stats](https://www.pullreview.com/github/yast/yast-registration/reviews/refactoring)

ancorgs:
One thing I forgot to mention during previous review. The tests on this module are not compatible with RSpec 3 (we haven't notice because yast-registration is not in openSUSE). We should use the opportunity of the refactoring to remove all the ```be_true``` and ```be_false```.

ancorgs:
Speaking about ```be_true```, this check is confusing.
https://github.com/yast/yast-registration/blob/master/test/registration_ui_test.rb#L36

The call to ```register_system_and_base_product``` returns ```[true, []]``` which matches ```be_true``` but is not equal to ```true```.

It's not exactly something broken by this PR, is already wrong (IMO) in master, but I noticed it while reviewing the tests on this PR.

ancorgs:
I guess we could keep discussing tests and code architecture forever, but this is already a big step forward, so LGTM.

lslezak:
@ancorgs Thanks for the review and help with the tests! :+1:

PR #186, opened 2015-04-24
Allow network setup in installed system

lslezak:
- Allow network configuration also in installed system (network configuration is required in https://trello.com/c/GGlxEy3u)
- 3.1.130

### DIFF
--- a/src/lib/registration/helpers.rb
+++ b/src/lib/registration/helpers.rb
@@ -85,9 +85,16 @@ module Registration
       (descr && !descr.empty?) ? "#{descr} (#{url})" : url
     end
 
+    # run the network configuration module
     def self.run_network_configuration
       log.info "Running network configuration..."
       Yast::WFM.call("inst_lan", [{ "skip_detection" => true }])
+    end
+
+    # check if the network configuration module is present
+    # @return [Boolean] true if the network configuration client is present
+    def self.network_configurable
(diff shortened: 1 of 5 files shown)


src/lib/registration/helpers.rb
```diff
@@ -85,11 +85,18 @@ def self.service_description(service)
...
+    # run the network configuration module
     def self.run_network_configuration
       log.info "Running network configuration..."
       Yast::WFM.call("inst_lan", [{ "skip_detection" => true }])
     end
 
+    # check if the network configuration module is present
+    # @return [Boolean] true if the network configuration client is present
+    def self.network_configurable
+      Yast::WFM.ClientExists("inst_lan")
```
ancorgs: Is this false for installed systems? I have inst_lan.rb in my system

Or is simply safe to call that client in an already installed system?

lslezak: It depends, by default it's installed. The check makes sure the client (and thus the `yast2-network`) package is installed. In installation we are pretty sure it's present, but in installed system user might remove it.

This check just guards the `Yast::WFM.call("inst_lan")` call to avoid possible errors.

ancorgs:
I haven't notice the "rescue". LGTM

lslezak:
Thanks!

PR #187, opened 2015-05-14
Support for re-registering extensions and modules (FATE#318800)

lslezak:
Note: There is still one part missing: the old subscription should be "unlinked" before registering an extension again. But this requires an extension in the `connect` gem which won't be ready in this sprint, I'll add it later when it's ready.

### DIFF
--- a/src/lib/registration/ui/addon_selection_base_dialog.rb
+++ b/src/lib/registration/ui/addon_selection_base_dialog.rb
@@ -6,7 +6,7 @@ require "registration/addon_sorter"
 module Registration
   module UI
     # this class displays and runs the dialog with addon selection
-    class AddonSelectionDialog
+    class AddonSelectionBaseDialog
       include Yast::Logger
       include Yast::I18n
       include Yast::UIShortcuts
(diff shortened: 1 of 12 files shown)


mchf:
Travis fails

src/lib/registration/ui/addon_selection_base_dialog.rb
```diff
@@ -39,45 +31,33 @@ def initialize(registration)
...
-          # help text (3/3)
-          _("<p>If you want to remove any extension or module you need to log"\
-              "into the SUSE Customer Center and remove them manually there.</p>"),
-          # always enable Back/Next, the dialog cannot be the first in workflow
-          true,
-          true
-        )
-
-        @old_selection = Addon.selected.dup
+        raise "Not implented"
```
mchf: typo

lslezak:
Travis is OK, just coveralls fails because of the 0.31% test coverage decrease...

src/lib/registration/ui/addon_selection_reregistration_dialog.rb
```diff
@@ -0,0 +1,73 @@
...
+          # help text (2/3)
+          _("<p>Please note, that some extensions or modules might need "\
+              "specific registration code.</p>") +
+          # help text (3/3)
+          _("<p>If you want to remove any extension or module you need to log"\
+              "into the SUSE Customer Center and remove them manually there.</p>"),
+          # always enable Back/Next, the dialog cannot be the first in workflow
+          true,
+          true
+        )
```
mchf: texts are same as in the other class, may be it deserve an optimization.

mchf:
LGTM for me

PR #188, opened 2015-06-05
Rename "SUSE Customer Center Configuration" to "Product Registration" (FATE#318800)

mvidner:
[FATE#318800](https://fate.suse.com/318800)

This is only a proposal, wait with merging until PM confirm the name.


src/clients/registration.rb
```diff
@@ -0,0 +1 @@
+scc.rb
```
lslezak: Wow, a nice trick how to redirect to a different client...

lslezak:
LGTM, so let's wait for the PM's feedback...

mvidner:
@jsrain re-renamed

jsrain:
LGTM, provided the pending check passes

PR #189, opened 2015-06-09
Renamed the module again to "Product Registration"

mvidner:
after a clarification in FATE#318800.

joe+sarsene, https://fate.suse.com/318800#comment_31


ancorgs:
LGTM

PR #190, opened 2015-06-16
adapt to new suse connect

jreidinger:


lslezak:
Please update also the dependency in the `*.spec` file.

lslezak:
The connect version is frozen in https://github.com/yast/yast-registration/blob/master/Gemfile#L4, please update it.

lslezak:
LGTM, thanks! :+1:

PR #192, opened 2015-07-02
use suse-connect master branch at Travis

lslezak:
Build against the `master` suse-connect, according to the SCC team this will go to the SLE12-SP1 and will be also released as a maintenace update for SLE12 (needed for online migration).


jreidinger:
LGTM

PR #193, opened 2015-07-07
Use "zypper dup --no-vendor-change" equivalent

lslezak:
- Use `Pkg.PkgUpdateAll()` instead of `Pkg.AddUpgradeRepo()`
- Disable the vendor change solver flag
- 3.1.136

### DIFF
--- a/src/lib/registration/migration_repositories.rb
+++ b/src/lib/registration/migration_repositories.rb
@@ -17,67 +17,88 @@ require "yast"
 require "registration/sw_mgmt"
 
 module Registration
-  # this class activates the migration repositories
+  # this class activates the migration services and repositories
   class MigrationRepositories
     include Yast::Logger
 
     Yast.import "Pkg"
 
+    # reset the libzypp migration setup
     def self.reset
-      log.info "Resetting upgrade repos config"
-      repos = Yast::Pkg.GetUpgradeRepos()
-      repos.each { |repo| Yast::Pkg.RemoveUpgradeRepo(repo) }
+      # reset the solver
+      Yast::Pkg.SetSolverFlags("reset" => true)
 
       # deselect all pre-selected packages or patches
       Yast::Pkg.PkgReset
     end
 
-    attr_accessor :repositories, :install_updates
+    attr_accessor :repositories, :services, :install_updates
 
     def initialize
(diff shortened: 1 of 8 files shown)


src/lib/registration/migration_repositories.rb
```diff
@@ -17,67 +17,90 @@
...
 
-    # use the repositories from a libzypp service for the migration
-    # @param [String] service_name name of the service
-    def add_service(service_name)
-      service_repos = SwMgmt.service_repos(service_name)
-      repositories.concat(service_repos)
+    # does any configured service contain an update repo?
+    # @return [Boolean] true if at least one service repository is an update
+    #   repository
+    def service_update_repo?
```
jreidinger: I miss here some glue betwee service and update repo like `service_with_update_repo?`?

src/lib/registration/migration_repositories.rb
```diff
@@ -17,67 +17,90 @@
...
+      all_repos.each do |repo|
+        repo_data = Yast::Pkg.SourceGeneralData(repo)
+
+        # enabled migration repositories, disable the others, change
+        # the status if it is different than expected
+        next if repositories.include?(repo) == repo_data["enabled"]
+
+        new_state_msg = repo_data["enabled"] ? "Disabling" : "Enabling"
+        log.info "#{new_state_msg} repository #{repo_data["alias"]}"
+        Yast::Pkg.SourceSetEnabled(repo, !repo_data["enabled"])
```
jreidinger: why `!` for enabled??? very confusing :confused:

src/lib/registration/migration_repositories.rb
lslezak: A service contains repositories, this function returns `true` if any service provides an update repository, i.e. at least one update repository is provided by the services.

jreidinger: yeah, I get it from code, but method name is not much communicative...if I see below `MigrationRepositories.service_update_repo?` then it is hard to quess what it do

lslezak: I see... OK, I'll rename it, `service_with_update_repo?` sounds better.

jreidinger:
LGTM

PR #194, opened 2015-07-10
Set the correct vendor change flag (FATE#319128)

lslezak:
for the DUP mode instead of the normal mode

- 3.1.137


jreidinger:
LGTM, but how it behave if it is not defined for older versions?

lslezak:
The `yast2-pkg-bindings` dependency has been updated, it cannot be undefined.

jreidinger:
I mean your new condition in code

lslezak:
The target is SLE12-SP1 and SLE12-GA backport, other older versions from YaST:Head do not make sense for registration anyway.

PR #196, opened 2015-08-03
Migration UI improvements

lslezak:
This is the result after discussing the UI with Ken. The changes include:

### The migration selection dialog:

- Sort the products in the displayed migration (the base product is at the end)
- Hide architecture in product name (not useful, architecture cannot be changed during migration anyway)
- Display user-friendly status

### Manual repository selection:

- Hide the repository URL in the `SelectionBox` (too long...)
- Display it in the details below (with some more data like the repository priority)

### Other changes:

- Query the registered extensions/modules and add them to the migration
  (to migrate also the registrations for the uninstalled products), this was required by the SCC team

### DIFF
--- a/src/lib/registration/addon.rb
+++ b/src/lib/registration/addon.rb
@@ -59,6 +59,8 @@ module Registration
         @selected ||= []
       end
 
+      # return add-ons which are registered but not installed in the system
+      # @return [Array<Addon>] the list of add-ons
       def registered_not_installed
         registered.select do |addon|
           !SwMgmt.installed_products.find do |product|
--- a/src/lib/registration/addon.rb
+++ b/src/lib/registration/addon.rb
@@ -20,6 +20,7 @@
 #
 
 require "forwardable"
+require "registration/sw_mgmt"
 
 module Registration
   # this is a wrapper class around SUSE::Connect::Product object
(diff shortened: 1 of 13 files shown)


src/lib/registration/addon.rb
```diff
@@ -58,6 +59,16 @@ def selected
         @selected ||= []
       end
```
schubi2: I would add here a little headline like : returns products which are registered but not installed.

src/lib/registration/ui/migration_repos_selection_dialog.rb
```diff
@@ -48,17 +48,25 @@ def run
...
           true,
           true
         )
 
+        update_repo_details
+
         loop do
           ret = Yast::UI.UserInput
 
-          store_values if ret == :next
```
schubi2: nice cleanup !

test/migration_repos_workflow_spec.rb
```diff
@@ -39,7 +39,7 @@
 
       let(:set_success_expectations) do
```
imobach: It's not related with the PR, but it does not seem to be a correct use case for `let`. Writing a helper method will be a better solution. But, as I said, it's out of the scope of this PR.

src/lib/registration/ui/migration_selection_dialog.rb
```diff
@@ -134,14 +144,51 @@ def update_details
...
+        details = sorted_migrations[idx].map do |product|
+          installed = installed_products.find do |installed_product|
+            installed_product["name"] == product.identifier
+          end
+
+          "<li>" + product_summary(product, installed) + "</li>"
         end
 
         details = "<ul>" + details.join + "</ul>"
-        _("<h3>Migration Products Details</h3>%s") % details
```
schubi2: Hm, I would split this, because the translation team should not take care about the html tags.

schubi2:
Just small comments. It is up to you to change it. LGTM. nice code :+1:

schubi2:
LGTM again :-)

lslezak:
@schubi2 thanks! :+1:

PR #197, opened 2015-08-07
Allow to go back during installation (bnc#940915)

ancorgs:


imobach:
LGTM

mvidner:
I just love hairy workflows. Anyway,  see the comment "# skip this when going back"? It points to code that was deleted in
https://github.com/ancorgs/yast-registration/commit/3132f777812015941ff5bd91ac4586afe2407cd4

ancorgs:
@mvidner I completely overlooked that comment! Bringing that code back will have the same effect than this PR?

mvidner:
Well, I don't know for sure. A `true` flag on an "alias" marks it as `WS_special` for this code: https://github.com/yast/yast-yast2/blob/ad8fdd00fd7eb740cb8498b8f491dc96436f63b7/library/sequencer/src/modules/Sequencer.rb#L351 which means that a `:back` will not return to its dialog but act as if it returned `:back`.

Off on a tangent, I see a nice possibility for improvement for the Sequencer library where it would receive some data and be able to *draw* the workflow (with graphviz?) in addition to performing it. The first step would be allowing methods as symbols instead of lambdas in

```rb
aliases = {
 "check"                  => ->() { registration_check },  # before
 "check"                  => :registration_check           # after
```

PR #199, opened 2015-08-14
Catch exceptions also when loading the available extensions (bsc#941491)

lslezak:
- 3.1.141

Do not crash when an exception is raised.

(Note: An error popup is displayed internally inside the `ConnectHelpers.catch_registration_errors` it's enough to just check the return value).

### DIFF
--- a/test/inst_scc_test.rb
+++ b/test/inst_scc_test.rb
@@ -28,6 +28,16 @@ describe "inst_scc client" do
       expect(Yast::WFM.call("inst_scc")).to eq(:abort)
     end
 
+    it "displays an error when loading the available extensions fails" do
+      # click installing extensions, close the error message and abort the workflow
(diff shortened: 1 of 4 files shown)


ancorgs:
Nothing added to `inst_scc_test.rb`? :wink:

test/inst_scc_test.rb
```diff
@@ -28,6 +28,16 @@
       expect(Yast::WFM.call("inst_scc")).to eq(:abort)
     end
 
+    it "displays an error when loading the available extensions fails" do
```
ancorgs: Well, strictly speaking you are not checking that the error is displayed. You are simply checking that the whole thing does not explode... which is actually nice to know (and proves the bug to be fixed) but not what the predicate says. :-)

ancorgs:
LGTM

lslezak:
Thanks, let's merge #200 first...

PR #200, opened 2015-08-14
Do not always mention installation (bnc#941403, bnc#941739)

ancorgs:
I have not even run this code, but since I'm having problems to get my SP1 updated/working...

### DIFF
--- a/src/lib/registration/ui/abort_confirmation.rb
+++ b/src/lib/registration/ui/abort_confirmation.rb
@@ -1,0 +1,44 @@
+require "yast"
+
+module Registration
+  module UI
+    # this class displays a popup for the user to confirm the abort
+    class AbortConfirmation
+      include Yast::I18n
+
+      Yast.import "Popup"
+      Yast.import "Mode"
+      Yast.import "Label"
+
+      # displays a new popup
+      # @return [Boolean] true whether the user confirmed he/she wants to abort
+      def self.run
+        dialog = new
+        dialog.run
+      end
+
+      # the constructor
+      def initialize
+        textdomain "registration"
+      end
+
+      # displays the popup
+      # @return [Boolean] true whether the user confirmed he/she wants to abort
+      def run
+        if Yast::Mode.installation
+          Yast::Popup.ConfirmAbort(:painless)
+        else
+          # Use .AnyQuestion instead of .YesNo or .ReallyAbort to have full control
+          # on :focus_no and be consistent with the .ConfirmAbort behavior
+          Yast::Popup.AnyQuestion(
+            Yast::Popup::NoHeadline(),
+            _("Really abort?"),
+            Yast::Label.YesButton,
+            Yast::Label.NoButton,
+            :focus_no
+          )
+        end
+      end
+    end
+  end
+end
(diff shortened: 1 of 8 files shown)


src/lib/registration/ui/abort_confirmation.rb
```diff
@@ -0,0 +1,42 @@
...
+
+      # the constructor
+      def initialize
+        textdomain "registration"
+      end
+
+      # displays the popup
+      # @return [Boolean] true whether the user confirmed he/she wants to abort
+      def run
+        if Yast::Mode.installation || Yast::Mode.repair
```
lslezak: The `repair` mode is obsolete, it's not used anymore (the repair module has been dropped).

src/lib/registration/ui/abort_confirmation.rb
```diff
@@ -0,0 +1,42 @@
...
+          Yast::Popup.ConfirmAbort(:painless)
+        else
+          msg = _("Really abort the registration?\n\n" \
```
lslezak: I'd probably prefer simple "Rally abort?" question with `[Yes][No]` buttons.

- The `no changes will be done to the system` statement is not completely true, registration workflow has several steps, e.g. you register the base product, then can optionally register and install an extension. But when you abort the extension selection dialog, the system has been already registered (=changed)

- `Really abort the *registration*?` is also not completely correct, some registration parts are called from the online migration module, distinguishing regular registration or online migration would make it complex.

Therefore just the simple "Rally abort?" is the best IMHO...

ancorgs: Ok. I was completely unsure about all that, so I decided to "clone" the installation abort message and let you spot the inconsistencies. I'll simplify the text.

src/lib/registration/ui/abort_confirmation.rb
```diff
@@ -0,0 +1,44 @@
...
+      end
+
+      # displays the popup
+      # @return [Boolean] true whether the user confirmed he/she wants to abort
+      def run
+        if Yast::Mode.installation
+          Yast::Popup.ConfirmAbort(:painless)
+        else
+          # Use .AnyQuestion instead of .YesNo or .ReallyAbort to have full control
+          # on :focus_no and be consistent with the .ConfirmAbort behavior
```
lslezak: Off-topic: Ideally we should add a new generic method for popups and take advantage of optional parameters in Ruby, something like `Popup.show(message: "msg", default_button: :no, header: "label",...)`...

lslezak:
Could you add a (simple) test case?

src/lib/registration/ui/abort_confirmation.rb
ancorgs: Absolutely. We have a lot of methods in `Popup` that are different on each other just by subtle and silly small differences. The API looks completely weird and it took me a while to figure out the best way to imitate `ConfirmAbort`.

ancorgs:
Good point. It should be fairly easy to add one.

lslezak:
LGTM, thanks a lot! :+1:

PR #202, opened 2015-08-21
Release target fix

lslezak:
- Tested registering the SLES-SP1 POOL product against SCC and SMT - both OK.
- It's the same as the SUSEConnect implementation: https://github.com/SUSE/connect/blob/master/lib/suse/connect/zypper/product.rb#L18.
- Fixed a small issue in the addon selection dialog, using just `addon.identifier` as a widget ID is not enough, there might be the same addon in more versions. In that case the ID is not unique and that makes troubles (found in the SMT testcase).

### DIFF
--- a/src/lib/registration/sw_mgmt.rb
+++ b/src/lib/registration/sw_mgmt.rb
@@ -50,8 +50,11 @@ module Registration
 
     ZYPP_DIR = "/etc/zypp"
 
-    FAKE_BASE_PRODUCT = { "name" => "SLES", "arch" => "x86_64", "version" => "12",
-      "release_type" => "DVD", "version_version" => "12" }
+    FAKE_BASE_PRODUCT = { "name" => "SLES", "arch" => "x86_64", "version" => "12-0",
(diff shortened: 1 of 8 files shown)


src/lib/registration/sw_mgmt.rb
```diff
@@ -449,6 +450,20 @@ def self.products_from_repo(repo_id)
...
     end
 
-    private_class_method :each_repo
+    def self.get_release_type(product)
+      if product["product_line"]
+        oem_file = File.join(OEM_DIR, product["product_line"])
+
+        if File.exist?(oem_file)
+          # read only the first line
+          line = File.open(oem_file, &:readline)
```
jreidinger: nice

jreidinger:
LGTM

src/lib/registration/sw_mgmt.rb
lslezak: Yeah, google and stackoverflow helped much here... :wink:

PR #204, opened 2015-08-26
Better popup question

lslezak:
- Use a better wording (suggested by Ken, see https://bugzilla.suse.com/show_bug.cgi?id=942843#c5)
- Added a message for later (that feature is still not clear, but make sure we have the text prepared before the translation deadline, just in case...)

### DIFF
--- a/src/lib/registration/ui/migration_repos_workflow.rb
+++ b/src/lib/registration/ui/migration_repos_workflow.rb
@@ -47,8 +47,8 @@ module Registration
         # before the translation deadline...
         # TRANSLATORS: popup question, confirm installing the available
         # updates now
-        _("There are some online updates available to installation,\n" \
-            "it is recommended to install all updates before proceeding.\n\n" \
+        _("Online updates are available for installation.\n" \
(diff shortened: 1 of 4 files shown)


src/lib/registration/ui/migration_repos_workflow.rb
```diff
@@ -43,6 +43,14 @@ def self.run
       def initialize
         textdomain "registration"
 
+        # a dummy message which will be used later, just make sure we have it
+        # before the translation deadline...
+        # TRANSLATORS: popup question, confirm installing the available
+        # updates now
+        _("There are some online updates available to installation,\n" \
```
jsrain: My English may not be best, anyway, here is my suggestion:
There are online update pending for installation.
[...]
Install pending updates now?

mvidner: I'd use "for" and a period:
```rb
_("Online updates are available for installation.\n" \
"It is recommended to install all updates before proceeding.\n\n" \
"Would you like to install the updates now?")
```

mvidner:
LGTM

PR #205, opened 2015-08-28
Keep the original NCCCredentials file permissions (bsc#943568)

lslezak:
...when upgrading from SLE11

- 3.1.146

(See http://ruby-doc.org/stdlib-2.2.0/libdoc/fileutils/rdoc/FileUtils.html#method-c-cp for the details in the documentation.)

### DIFF
--- a/test/sw_mgmt_spec.rb
+++ b/test/sw_mgmt_spec.rb
@@ -179,7 +179,7 @@ describe Registration::SwMgmt do
         .and_return(false)
 
       expect(FileUtils).to receive(:cp).with(File.join(root_dir, target_dir, "NCCcredentials"),
-        File.join(target_dir, "SCCcredentials"))
+        File.join(target_dir, "SCCcredentials"), preserve: true)
       expect(SUSE::Connect::Credentials).to receive(:read)
 
       expect { subject.copy_old_credentials(root_dir) }.to_not raise_error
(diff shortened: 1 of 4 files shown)


jsrain:
LGTM, provided that the pending checks succeed

test/sw_mgmt_spec.rb
```diff
@@ -178,8 +178,8 @@
       expect(File).to receive(:exist?).with(File.join(root_dir, target_dir, "SCCcredentials"))
         .and_return(false)
 
-      expect(FileUtils).to receive(:cp).with(File.join(root_dir, target_dir, "NCCcredentials"),
-        File.join(target_dir, "SCCcredentials"))
+      expect(subject).to receive(:`).with("cp -a " + File.join(root_dir, target_dir,
+        "NCCcredentials") + " " + File.join(target_dir, "SCCcredentials"))
```
jsrain: i assume that it will work correctly since it worked before, but still for my cusiousity: Why is root_dir joined only for the source file?

lslezak: Oh, found it: The file is copied from `/mnt/etc/...` (from the target) to `/etc` in the inst-sys. There it is found by SUSEConnect, later (in the `*_finish` script) it is copied to the target system.

That's the reason for the asymmetry in the target directory...

kobliha: Hmm, sound like something to document. At least in the code?

PR #206, opened 2015-09-01
Dynamic migration selection widget size (bsc#943636)

lslezak:
Additional improvements:

- SCC now provides the `product_type` flag, we can use the same sorter for sorting the addons in the migration target as in the addon selection dialog
- Use the new Connect API for the credentials management


imobach:
LGTM

PR #207, opened 2015-09-02
fixed syntax error (bsc#944089)

jsrain:


imobach:
As the objetive of the PR was to fix the '/', LGTM.

lslezak:
@imobach `<<` does not work for translated strings.

`_()` returns a frozen string (cannot be modified by `<<`). Of course, for normal `String` objects `<<` should be preferred.

imobach:
@lslezak Oh, I see. Thanks for the hint.

PR #208, opened 2015-09-07
Restore the *-Update repositories at the end of migration

lslezak:
After migration with disabled updates (migrating to the original SP release) the Update repositories need to be re-enabled.

https://bugzilla.suse.com/show_bug.cgi?id=943960

### DIFF
--- a/src/lib/registration/sw_mgmt.rb
+++ b/src/lib/registration/sw_mgmt.rb
@@ -62,16 +62,16 @@ module Registration
     def self.init(load_packages = false)
       # false = do not allow continuing without the libzypp lock
       lock = PackageLock.Connect(false)
-      raise pkg_exception unless lock["connected"]
+      raise_pkg_exception unless lock["connected"]
 
       # display progress when refreshing repositories
       PackageCallbacks.InitPackageCallbacks
 
-      raise pkg_exception unless Pkg.TargetInitialize(Installation.destdir)
-      raise pkg_exception unless Pkg.TargetLoad
-      raise pkg_exception unless Pkg.SourceRestore
+      raise_pkg_exception unless Pkg.TargetInitialize(Installation.destdir)
+      raise_pkg_exception unless Pkg.TargetLoad
+      raise_pkg_exception unless Pkg.SourceRestore
 
-      raise pkg_exception if load_packages && !Pkg.SourceLoad
+      raise_pkg_exception if load_packages && !Pkg.SourceLoad
     end
 
     # during installation /etc/zypp directory is not writable (mounted on
@@ -473,8 +473,8 @@ module Registration
       product["register_release"]
     end
 
-    def self.pkg_exception
-      PkgError.new(Pkg.LastError)
+    def self.raise_pkg_exception
+      raise PkgError.new, Pkg.LastError
     end
 
     private_class_method :each_repo, :get_release_type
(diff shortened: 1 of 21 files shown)


lslezak:
JFY: This is only the first part, a small change in `yast2-migration` package will follow...

src/lib/registration/sw_mgmt.rb
```diff
@@ -67,6 +69,7 @@ def self.init
       Pkg.TargetInitialize(Installation.destdir)
       Pkg.TargetLoad
       Pkg.SourceRestore
+      load_packages ? Pkg.SourceLoad : true
```
jreidinger: why return true? return state is not documented

src/lib/registration/ui/migration_finish_workflow.rb
```diff
@@ -0,0 +1,91 @@
...
+          ret = run_sequence
+        rescue => e
+          log.error "Caught error: #{e.class}: #{e.message.inspect}, #{e.backtrace}"
+          # TRANSLATORS: error message, %s are details
+          Yast::Report.Error(_("Internal error: %s") % e.message)
+          ret = :abort
+        end
+
+        ret
+      end
```
jreidinger: ret at begin is not needed:

```
def run
  run_sequence
rescue => e
  log.error "Caught error: #{e.class}: #{e.message.inspect}, #{e.backtrace}"
  # TRANSLATORS: error message, %s are details
  Yast::Report.Error(_("Internal error: %s") % e.message)
   return :abort
end
```

src/lib/registration/ui/migration_finish_workflow.rb
```diff
@@ -0,0 +1,91 @@
...
+
+      # run the workflow
+      # @return [Symbol] the UI symbol
+      def run_sequence
+        aliases = {
+          "restore_repos" => ->() { restore_repos }
+        }
+
+        ui = Yast::Sequencer.Run(aliases, WORKFLOW_SEQUENCE)
+        log.info "User input: #{ui}"
```
jreidinger: is it really user input? it looks like result of script and BTW from below lines I do not see how it can return something else then `:next`

src/lib/registration/ui/migration_repos_workflow.rb
```diff
@@ -256,8 +257,8 @@ def register_migration_products
         # synchronize the changes done by modifying the services,
         # reinitialize the repositories and reload the available packages
         Yast::Pkg.SourceFinishAll
-        Yast::Pkg.SourceRestore
-        Yast::Pkg.SourceLoad
+        Yast::Pkg.TargetFinish
+        SwMgmt.init(true)
```
jreidinger: return value ignore, but not above. Is it correct? looks strange

src/lib/registration/ui/migration_repos_workflow.rb
```diff
@@ -153,14 +157,11 @@ def run_sequence
       def find_products
         log.info "Loading installed products"
 
-        if !SwMgmt.init
+        if !SwMgmt.init(true)
```
jreidinger: I prefer here to use exception, so problem with ignored return value below cannot happen

src/lib/registration/ui/migration_repos_workflow.rb
lslezak: Fixed, it now throws an exception.

jreidinger:
just NP, otherwise LGTM

jreidinger:
LGTM

PR #210, opened 2015-09-10
Set the $releasever URL variable to the new base product (bsc#941563)

lslezak:
https://bugzilla.suse.com/show_bug.cgi?id=941563

### DIFF
--- a/src/lib/registration/releasever.rb
+++ b/src/lib/registration/releasever.rb
@@ -1,0 +1,96 @@
+# ------------------------------------------------------------------------------
+# Copyright (c) 2015 SUSE LLC, All Rights Reserved.
+#
+# This program is free software; you can redistribute it and/or modify it under
+# the terms of version 2 of the GNU General Public License as
+# published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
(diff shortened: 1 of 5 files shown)


src/lib/registration/releasever.rb
```diff
@@ -0,0 +1,96 @@
...
+    # @return [Array<Fixnum>] the repository list
+    def repos_to_refresh
+      repos = Yast::Pkg.SourceGetCurrent(true)
+      repos.select! do |repo|
+        # check the raw URL (without expanded variables)
+        raw_url = Yast::Pkg.SourceGeneralData(repo)["raw_url"]
+
+        next false unless raw_url
+
+        # The URL variables can be encoded is several ways, like $releasever or
```
imobach: I'm not sure if this documentation will be available in the one generated by yardoc. I would be nice to have it available at class-level (or at least at method level). Just a suggestion.

imobach:
Just a minor suggestion (feel free to ignore it). LGTM.

src/lib/registration/releasever.rb
lslezak: I don't think it's needed in the generated yardoc documentation, it's an internal description of the implementation. Moreover the method is private, it's not included in the yardoc output abyway.

It's there just to describe why the simple generic `$` check is used instead of something more specific like `$releasever`.

imobach: I was referring to the first and second paragraphs. But it's fine with me :)

PR #212, opened 2015-09-15
Update RSpec config

lslezak:
- the expect syntax is default in RSpec3 (no need to set it)
- enable verifying doubles, remove the obsolete (invalid) mocks
- use RSpec3 at Travis


jreidinger:
LGTM

PR #213, opened 2015-09-16
Better handle repeated SSL certificate import

lslezak:


src/lib/registration/connect_helpers.rb
```diff
@@ -215,6 +215,20 @@ def self.ask_import_ssl_certificate(cert)
...
+      if Storage::InstallationOptions.instance.imported_cert_sha256_fingerprint
+        # TRANSLATORS: multiline error message - a SSL certificate has been
+        # imported but the registration server still cannot be accessed securely,
+        # user has to solve the certificate issue manually.
+        Yast::Report.Error(_("A certificate has been already imported\n" \
+          "but the server connection still cannot be trusted.\n\n" \
+          "Please fix the certificate issue manually, ensure that the server\n" \
+          "can be connected securely and start the YaST module again."))
+
+        return false
```
mvidner: It seems useful to document the return value, because we've made a subtle change.
Before:
> \# @return [Boolean] certificate successfully imported

Now:
> \# @return [Boolean] success? can fail if cannot import or the cert is not valid after all

mvidner:
Thanks, LGTM!

PR #214, opened 2015-09-22
Rollback registration when online migration is aborted

lslezak:
- Added a new client `registration_sync_workflow.rb` for registration rollback (will be called from the `yast2-migration` module)
- Refactored the clients (moved the code to the library for easier testing, added a base class to share the common code) so the CodeClimate score [is kept roughly the same](https://codeclimate.com/github/yast/yast-registration/compare/rollback#ratings) (otherwise it would decrease).
- Some more small fixes
- Increased code coverage


imobach:
LGTM

PR #217, opened 2015-09-25
Do not crash when a repository cannot be accessed (bsc#946200)

lslezak:
https://bugzilla.suse.com/show_bug.cgi?id=946200

Tested in online migration with SLE-WE extension.

### DIFF
--- a/test/sw_mgmt_spec.rb
+++ b/test/sw_mgmt_spec.rb
@@ -367,4 +367,69 @@ describe Registration::SwMgmt do
       subject.zypp_config_writable!
     end
   end
+
+  describe ".check_repositories" do
+    let(:repo) { 42 }
+
+    before do
+      allow(Yast::Pkg).to receive(:SourceGetCurrent).with(true).and_return([repo])
+      allow(Yast::Pkg).to receive(:SourceGeneralData).with(repo)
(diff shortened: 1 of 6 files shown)


test/sw_mgmt_spec.rb
```diff
@@ -367,4 +367,69 @@
...
+    it "returns true if all repositores refresh" do
+      expect(Yast::Pkg).to receive(:SourceRefreshNow).with(repo).and_return(true)
+
+      expect(subject.check_repositories).to eq(true)
+    end
+
+    context "a repository refresh fails" do
+      before do
+        expect(Yast::Pkg).to receive(:SourceRefreshNow).with(repo).and_return(false)
+        allow(Yast::Popup).to receive(:ErrorAnyQuestion).and_return(false)
```
imobach: Is this line needed? It looks like you set an 'expect' in each test.

imobach:
Just a minor comment. Otherwise, LGTM.

test/sw_mgmt_spec.rb
lslezak: Good point, I wanted to share this code, but in the end I used a different mock in each test and this line become useless. I'll remove it.

PR #220, opened 2015-10-09
Addon selection dialog - sort the addons by label (bsc#949424)

lslezak:
### One more fix for the urgent maintenance update for SES

Sorting by `name` does not work as expected, the name might not be unique (see https://bugzilla.suse.com/show_bug.cgi?id=949424#c13).

Moreover `label` is actually displayed in the UI so it should be used for sorting as well.

The testing data file (`fixtures/available_addons.yml`) has been updated by the current SCC response.


imobach:
LGTM

jsrain:
@imobach You were faster :-)

lslezak:
@imobach @jsrain  thanks to both of you :wink:

PR #222, opened 2015-10-09
Update the Connect version dependency

lslezak:
This is a workaround for [bsc#949424](https://bugzilla.suse.com/show_bug.cgi?id=949424) in the original SLE-12-GA installation system. The bug is about conflict between the SES1 and SES2 extensions where only one of them is handled properly.

The connect version is sent in the request header, the workaround is that the SCC server will return only SES2 extension and will hide SES1 when it detects the registration from the SLE-12-GA installer.

The `yast2-registration` maintenance update will be released together with a new Connect where the only difference will be the version number (the other incompatible changes will be released later together with the migration patch). For this Connect version (actually all except the GA version) the SCC server will return both SES1 and SES2 extensions as the fixed YaST package can handle them correctly.


ancorgs:
Doesn't this need an entry in the changelog?

lslezak:
Um, yes, I forgot about it, thanks.

ancorgs:
LGTM

PR #223, opened 2015-10-12
[Review] Request from 'schubi2' @ 'yast/yast-registration/review_151012_registration_server_autoyast'

schubi2:
Please review the following changes:
  * 4474fca packaging
  * 7827c31 taking url from autoyast conifiguration file


src/lib/registration/url_helpers.rb
```diff
@@ -119,6 +120,13 @@ def self.reg_url_at_installation
       slp_service_url
     end
 
+    # get registration URL from AutoYaST configuration file
+    def self.reg_url_from_autoyast_config
+      server = ::Registration::Storage::Config.instance.reg_server
+      return server if server && !server.empty?
+      SUSE::Connect::Config::DEFAULT_CONFIG_FILE
```
lslezak: JFYI: YaST should only use the constants/methods from the [connect/yast.rb](https://github.com/SUSE/connect/blob/master/lib/suse/connect/yast.rb) file. The rest is considered unstable and can be changed by the SCC team anytime...

And the constant should be the actually the `DEFAULT_URL`, not the config file.

So the correct value is `SUSE::Connect::YaST::DEFAULT_URL`. Please adapt the tests as well.

src/lib/registration/url_helpers.rb
```diff
@@ -119,6 +120,13 @@ def self.reg_url_at_installation
       slp_service_url
     end
 
+    # get registration URL from AutoYaST configuration file
+    def self.reg_url_from_autoyast_config
+      server = ::Registration::Storage::Config.instance.reg_server
```
lslezak: Um, where is the `reg_server` value set during the AutoYaST installation? There is actually a [`set_registration_url` method defined in the scc_auto.rb client](https://github.com/yast/yast-registration/blob/5d614c6fb3787690ea377eb196e8c96d235f189d/src/clients/scc_auto.rb#L144) called at [`Write` here](https://github.com/yast/yast-registration/blob/5d614c6fb3787690ea377eb196e8c96d235f189d/src/clients/scc_auto.rb#L173).

schubi2: Yes, that's right, but while calling
Helpers::write_conifg  the function UrlHelpers.registration_url will be called and returns nil

2015-08-27 15:45:04 <2> 10.20.90.152(2926) [Ruby] registration/url_helpers.rb:73 Unknown mode: autoinstallation, using default URL
2015-08-27 15:45:04 <1> 10.20.90.152(2926) [Ruby] registration/helpers.rb:100 Boot reg_ssl_verify option: nil
2015-08-27 15:45:04 <1> 10.20.90.152(2926) [Ruby] registration/helpers.rb:134 writing registration config: {:url=>nil, :insecure=>false}

The interesting part is, that 
registration/url_helpers.rb:73 Unknown mode: autoinstallation, using default URL
will be logged only while calling the write_config function.

lslezak:
Ok, LGTM

@schubi2, thanks a lot  for fixing it!

PR #225, opened 2015-10-13
Disable the [Clone] button in AutoYaST (bsc#950098)

lslezak:
the registration module does not support cloning (yet)

- 3.1.160

See https://bugzilla.suse.com/show_bug.cgi?id=950098


ancorgs:
LGTM

PR #227, opened 2015-10-14
SLE12 backport - connect API update, migration support, small fixes

lslezak:
This is a pretty huge PR, but the vast majority of the patches are simply backports from `master` and were already reviewed, that means a quick scan should be enough.

- The main reason for this maintenance update is a planned maintenance update for the connect gem (the backend library) which changed the API and needs some changes in the YaST module.
- But the biggest part of the PR is in the online migration backport. The full online migration will be released later, but the registration part can be released sooner in this update to save some work later (for  us, the maintenance and the QA team).
- Additionally the PR contains some small bug fixes which would be nice to have also in GA (although it's questionable because for installing this update the system has to be already registered, but anyway...)
- And finally some changes are SLE12-GA specific (e.g. package version dependencies)


src/lib/registration/registration_ui.rb
```diff
@@ -181,8 +218,174 @@ def disable_update_repos(product_service)
...
+      # if registering only add-ons which do not need a reg. code (like SDK)
+      # then simply start the registration
+      if selected_addons.all?(&:free)
+        Yast::Wizard.SetContents(
+          # dialog title
+          _("Register Extensions and Modules"),
+          # display only the products which need a registration code
+          Empty(),
+          # help text
+          _("<p>Extensions and Modules are being registered.</p>"),
```
schubi2: html text should not be in the translation text. But anyway let it, if the translation already exists.

src/lib/registration/registration.rb
```diff
@@ -129,19 +148,32 @@ def activated_products
...
 
     private
 
     def set_registered(remote_product)
       addon = Addon.find_all(self).find do |a|
-        a.arch == remote_product.arch &&
-          a.identifier == remote_product.identifier &&
-          a.version  == remote_product.version &&
-          a.release_type == remote_product.release_type
+        a.matches_remote_product?(remote_product)
```
schubi2: nice cleanup !!

src/lib/registration/registration_ui.rb
```diff
@@ -181,8 +218,174 @@ def disable_update_repos(product_service)
...
+        ConnectHelpers.catch_registration_errors(
+          message_prefix: "#{product.label}\n") do
+          register_selected_addon(product, known_reg_codes[product.identifier])
+        end
+
+        # remove from selected after successful registration
+        selected_addons.reject! { |selected| selected.identifier == product.identifier }
+      end
+
+      !product_succeed.include?(false) # succeed only if noone failed
```
schubi2: typo ? :-)

src/lib/registration/registration_ui.rb
```diff
@@ -181,8 +218,174 @@ def disable_update_repos(product_service)
...
+          "reg_code" => reg_code,
+          "arch"     => product.arch,
+          "version"  => product.version
+        }
+
+        registration.register_product(product_data)
+      end
+
+      # select repositories to use in installation (e.g. enable/disable Updates)
+      select_repositories(product_service) if Yast::Mode.installation || Yast::Mode.update
```
schubi2: I have checked this. Yast::Mode.installation also includes AutoYaST case. So it is no problem....

schubi2:
The view comments which I have made are not worth for fixing (perhaps in master). Nice Code !!!!!
LGTM
A side node: There are a lot of new messages which are tagged for translation. Should we submit translations for update too ?

jsrain:
Since this is a back-port and the future development will happen in a different branch, the coverage decrease should IMO not block accepting this PR and submitting the package.

lslezak:
@schubi2 the new messages are mainly for online migration, the translations can be released later.

@jsrain yes, the code coverage drop was caused mainly by adding only the important commits from `master`. The code in `master` has more tests, but it uses RSpec3 and there was done some additional refactoring which would need more work for backporting so I simply skipped them .

src/lib/registration/registration_ui.rb
lslezak: Good point, but I'll keep it to not break the translations...

lslezak:
@schubi2 thanks for the review! :+1: Here is one virtual :beer: for you (or :wine_glass:, if you prefer) :wink:

schubi2:
Prost !!! :+1:

PR #228, opened 2015-10-19
Upgrade fix

lslezak:
Fixes [bsc#950795](https://bugzilla.suse.com/show_bug.cgi?id=950795)

Tested in a patched installer.

### DIFF
--- a/src/lib/registration/registration_ui.rb
+++ b/src/lib/registration/registration_ui.rb
@@ -118,9 +118,11 @@ module Registration
       [upgraded, product_service]
     end
 
+    # @param [Array<Registration::Addon>] addons to update
     def update_addons(addons, enable_updates: true)
       # find addon updates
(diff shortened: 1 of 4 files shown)


src/lib/registration/registration_ui.rb
```diff
@@ -337,6 +339,8 @@ def select_repositories(product_service)
       SwMgmt.set_repos_state(updates, install_updates?)
     end
 
+    # update addon registration to a new version
+    # @param [Array<Registration::Addon>] adddon addon to update
```
ancorgs: this should be `@param [Registration::Addon] addon...`

lslezak: Good point, copied from `update_addons` but forgot to remove the `Array`...

ancorgs:
lgtm

PR #230, opened 2015-11-02
Display a better error message when registration fails (bsc#952443)

lslezak:
...because of a typo in the URL

- 3.1.164

### Notes:

- When registration failed it was considered as an unconfigured network, but that's not always the case. If you use a SMT registration server and make a typo in the host name you would get the same error and that was confusing (the network really was configured).
- The move to a separate method was required by RuboCop, it started to complain about too long method... :wink: 
- The error message has been reused (because of the text freeze) and should be improved later, but there are also displayed the error details which should make it more clear what has happened...

#### The original state:

registration_invalid_hostname_bug

#### Fixed version:
registration_invalid_hostname_fixed

### DIFF
--- a/package/yast2-registration.changes
+++ b/package/yast2-registration.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Mon Nov  2 14:21:16 UTC 2015 - lslezak@suse.cz
+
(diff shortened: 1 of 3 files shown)


schubi2:
LGTM

PR #231, opened 2015-11-08
Always enable update repositories for modules (bsc#953536)

lslezak:
(during online migration)

- 3.1.165

Fixes [bsc#953536](https://bugzilla.suse.com/show_bug.cgi?id=953536)

### Bug Prerequisites

- Any module is registered in the system
- User selects to **not** install updates in this dialog:

migration_disabled_updates_question

### The Bug

The result is a dependency conflict in the migration summary:

migration_disabled_updates_broken_summary

There is a conflict in the package management:

migration_disabled_updates_broken_conflict

This is caused by bug [bsc#939702](https://bugzilla.suse.com/show_bug.cgi?id=939702).

All update repositories are disabled:

migration_disabled_updates_broken_repos

### Bug Fix

This patch fixes the problem:

migration_disabled_updates_fixed_summary

It always keeps the update repositories for modules enabled, the update repositories for the other product types (base product, extension) are diabled.

migration_disabled_updates_fixed_repos

### DIFF
--- a/package/yast2-registration.changes
+++ b/package/yast2-registration.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Fri Nov  6 20:26:06 UTC 2015 - lslezak@suse.cz
+
+- always enable update repositories for modules during online
+  migration (bsc#953536)
+- 3.1.165
+
+-------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


schubi2:
LGTM

PR #236, opened 2016-01-05
Fix wrong backport of bsc#941403

ancorgs:
We made an error backporting the fix to SLE12 and https://bugzilla.suse.com/show_bug.cgi?id=941403 got reopened. This should fix it.


lslezak:
LGTM

PR #241, opened 2016-02-08
Read registration codes from a USB stick

mvidner:
https://fate.suse.com/316796
https://trello.com/c/0WL8Ucks

## Registration Codes from a USB Stick

During the installation of a SUSE Linux Enterprise product, you are asked for a registration code. Previously you had to remember it and type it by hand. Now the code can be read from USB storage.

Use either `/regcodes.txt` or `/regcodes.txt` on a USB disk.

regcodes.txt:

```
SLES    cc36aae1
SLED    309105d4

sle-we  5eedd26a
sle-live-patching 8c541494
```

regcodes.xml

```xml
<?xml version="1.0"?>
<profile
 xmlns="http://www.suse.com/1.0/yast2ns"
 xmlns:config="http://www.suse.com/1.0/configns">
  <suse_register>
    <!-- See https://www.suse.com/documentation/sles-12/singlehtml/book_autoyast/book_autoyast.html#CreateProfile.Register.Extension -->
    <addons config:type="list">
      <addon>
        <!-- Name of add-on as listed by "zypper search --type product" -->
        <name>sle-we</name>
        <reg_code>5eedd26a</reg_code>
      </addon>
      <addon>
        <name>sle-live-patching</name>
        <reg_code>8c541494</reg_code>
      </addon>
      <addon>
        <!-- SLES is not an add-on but listing it here allows for combining
             several base product registration codes in a single file -->
        <name>SLES</name>
        <reg_code>cc36aae1</reg_code>
      </addon>
      <addon>
        <name>SLED</name>
        <reg_code>309105d4</reg_code>
      </addon>
    </addons>
  </suse_register>
</profile>
```


mvidner:
Now the prototype can read /regcodes.xml or /regcodes.txt from the USB stick, and the codes are prefilled for the addon products, but not yet for the base product.

src/lib/registration/storage.rb
```diff
@@ -30,9 +34,81 @@ module Storage
...
+      include Yast::Logger
 
       def initialize
-        self.reg_codes = {}
+        Yast.import "XML"
+        # TODO: package it better
+        initialize_autoinstall_io(self)
+
+        self.reg_codes = reg_codes_from_usb_stick || {}
+        log.info "Registration codes read for #{reg_codes.keys.inspect}"
```
lslezak: Good, we should not log the reg. codes, only the products. :+1:

lslezak:
Looks good so far, just refactor/clean up the `autoinstall_io.rb` file. :wink:

mvidner:
The Travis build fails because YaST:Head:Travis/yast2-update is not kept current (and I doubt it is worth fixing now)
I use `YAST_SUBMIT=sle12sp1 rake osc:build["-k /tmp -p /tmp"]` on both PRs in sequence. (The sp1 part is not normally necessary but right now Factory is making a warp jump)

mvidner:
Base product code now also works.

src/lib/registration/registration_codes_loader.rb
```diff
@@ -0,0 +1,100 @@
...
+      nil
+    end
+
+    # @param template for tempfile name
+    # @yieldparam actual file name
+    def with_tempfile(pattern, &block)
+      tempfile = Tempfile.new(pattern)
+      block.call(tempfile.path)
+    ensure
+      tempfile.unlink
```
lslezak: I'm not sure whether it's necessary, but the doc suggests to also call the `close` method as a good practice™. See http://ruby-doc.org/stdlib-2.2.0/libdoc/tempfile/rdoc/Tempfile.html#class-Tempfile-label-Explicit+close

src/lib/registration/registration_codes_loader.rb
```diff
@@ -0,0 +1,100 @@
...
+      return nil unless File.readable?(filename)
+      xml_hash = Yast::XML.XMLToYCPFile(filename)
+      parse_xml(xml_hash)
+    end
+
+    # @param xml_hash [Hash] as used in AY and returned by Yast::XML.XMLToYCPFile
+    # @return [Hash{String => String},nil]
+    def parse_xml(xml_hash)
+      suse_register = xml_hash.fetch("suse_register", {})
+      addons = suse_register.fetch("addons", [])
```
lslezak: This reads only the addon reg codes, the base product reg. code is stored at the `suse_register`/`reg_code` key, see [the schema definition](https://github.com/yast/yast-registration/blob/master/src/autoyast-rnc/registration.rnc#L10).

The problem is how to store the key as we do not know the base product name here... Use an empty string? See below.

src/lib/registration/ui/base_system_registration_dialog.rb
```diff
@@ -126,12 +126,19 @@ def product_details_widgets
       def input_widgets
         options = Storage::InstallationOptions.instance
 
+        reg_code = options.reg_code
+        if reg_code.empty?
+          known_reg_codes = Storage::RegCodes.instance.reg_codes
+          base_product_name = SwMgmt.find_base_product["name"]
+          reg_code = known_reg_codes[base_product_name] || ""
```
lslezak: :red_circle: Then here we could use `known_reg_codes[base_product_name] || known_reg_codes[""] || ""`.

src/lib/registration/registration_codes_loader.rb
```diff
@@ -0,0 +1,100 @@
...
+module Registration
+  # Aids Registration::Storage::RegCodes in loading the codes
+  # from external storage so that the user does not have to type them
+  # each time when doing multiple installations. FATE#316796
+  module RegCodesLoader
+    include Yast::I18n # missing in yast2-update
+    include Yast::Transfer::FileFromUrl
+
+    REGCODES_NAME_HANDLERS = {
+      "regcodes.xml" => :reg_codes_from_xml,
```
lslezak: We use the AY format for XML file, why not trying `autoyast.xml` file as well? With `autoyast.xml` on an USB stick you could either do a full autoinstall or pre-fill the reg. codes in standard installation. I think that would be a nice feature...

src/lib/registration/registration_codes_loader.rb
```diff
@@ -0,0 +1,100 @@
...
+require "yast"
+require "transfer/file_from_url"
+
+Yast.import "XML"
+
+module Registration
+  # Aids Registration::Storage::RegCodes in loading the codes
+  # from external storage so that the user does not have to type them
+  # each time when doing multiple installations. FATE#316796
+  module RegCodesLoader
```
lslezak: NTH: I think it would be nice to have some unit tests for the `RegCodesLoader` module with some testing data files in `test/fixtures`...

src/lib/registration/registration_codes_loader.rb
kobliha: Because having autoyast.xml available automatically starts AutoYast instead of the installer.

src/lib/registration/registration_codes_loader.rb
mvidner: I think using an empty string is a bad idea. It seems to make sense here in the code, but in regcodes.txt or regcodes.xml it does not work out. Not because of the syntax, but because IMHO you will want to have ONE such file with your current regcodes for multiple products, and that includes multiple base products.

Reading suse_register/reg_code , yes, I forgot that.

src/lib/registration/registration_codes_loader.rb
lslezak: Even on an USB stick? I thought it's true only for floppies...

mvidner: > having autoyast.xml available automatically starts AutoYast

Ah, indeed, as [documented](https://www.suse.com/documentation/sles-12/singlehtml/book_autoyast/book_autoyast.html#Commandline.ay) with `autoyast=default` (but I am not sure whether that is really the default).

Which reminds me, maybe we should allow, or even require, an `regcodes=(url)` argument to trigger this feature?

lslezak: IMHO no, `regcodes=(url)` argument would make it too complicated. Lets start with this simple automatic case and let see later if we get some feedback for improvements.

test/registration_codes_loader_spec.rb
```diff
@@ -0,0 +1,133 @@
...
+
+        expect(subject).to receive(:reg_codes_from_xml)
+          .with(tempfile)
+          .and_return("a" => "b")
+
+        expect(subject.reg_codes_from_usb_stick).to eq("a" => "b")
+      end
+    end
+
+    context "when only regcodes.txt exists" do
```
lslezak: And what if both exist? Which one is preferred? I think this case should be also covered so we still prefer the same. (And to detect a regression when the preferred file is changed.)

lslezak:
The [acceptance criteria](https://trello.com/c/0WL8Ucks) says _"installation-only feature"_ as far as I can see it does not check the current mode.

src/lib/registration/registration_codes_loader.rb
mvidner: In the end reading suse_register/reg_code is not implemented, and base product regcode is read from the addon section. Reading the other element would IMO complicate the code too much for little benefit. We can add later it if we get feedback.

test/registration_codes_loader_spec.rb
mvidner: Added a test case below.

mvidner:
"Installation-only" is now also checked.

lslezak:
LGTM

(Let's ignore the Travis failure for now, we can fix it later.)

mvidner:
Packages finally published, Travis passes: https://travis-ci.org/yast/yast-registration/builds/110980260

lslezak:
Great! :+1:

holgis:
Does YaST also understand the <version>-tag here? Like this:

    <addons config:type="list">
      <addon>
        <name>product</name>
        <reg_code>1234567890abcd</reg_code>
        <version>1.2</version>
      </addon>
    </addons>

That would be interesting while a product is still in development and uses a beta key.

kobliha:
@holgis The `version` does not seem to be in use in code. It's not in the documentation either.

holgis:
Are you sure? It seem to work for me when using SCC. Code snipped is from my autoyast file.

kobliha:
I'm not "sure", IMO it just ignores the version, but I might be wrong.

lslezak:
The AY profile actually supports `version` for addons (see https://github.com/yast/yast-registration/wiki/Autoyast-Support or the schema definition https://github.com/yast/yast-registration/blob/master/src/autoyast-rnc/registration.rnc#L19).

But here when reading from USB it is ignored. The question is whether we want to support it and make it more complex. Martin? 

IMHO using simpler approach is OK, you can always edit the file on the USB stick before starting installation to contain the reg. code for the desired version.

mvidner:
@holgis Sorry, I missed your question.
The `version` is ignored, and so is `arch`, and `release_type` (which I have no idea about what it is).
If you think it is worth it, please file a bug in bugzilla and we can work out the details there.

PR #242, opened 2016-02-12
Ask again after wrong registration code (bsc#958521)

teclator:
- We will register all valid addons and will continue asking for register codes for the missing ones. You can go back and disable the extension in case you decide to not install it.

### DIFF
--- a/package/yast2-registration.changes
+++ b/package/yast2-registration.changes
@@ -1,4 +1,12 @@
 -------------------------------------------------------------------
+Fri Feb 12 12:42:19 UTC 2016 - knut.anderssen@suse.com
+
+- Continue asking for registration codes also after failed
+  registration. you should go back to unselect some of not wanted
+  modules (bsc#958521)
+- 3.1.168
+
+-------------------------------------------------------------------
 Fri Jan 15 08:44:50 UTC 2016 - lslezak@suse.cz
 
 - Fixed a crash when EULA download fails (bsc#941232)
--- a/package/yast2-registration.changes
+++ b/package/yast2-registration.changes
@@ -1,9 +1,7 @@
 -------------------------------------------------------------------
 Fri Feb 12 12:42:19 UTC 2016 - knut.anderssen@suse.com
 
-- Continue asking for registration codes also after failed
-  registration. you should go back to unselect some of not wanted
-  modules (bsc#958521)
+- Ask again after some provided registration code failed (bsc#958521)
 - 3.1.168
 
 -------------------------------------------------------------------
--- a/src/clients/inst_scc.rb
+++ b/src/clients/inst_scc.rb
@@ -164,9 +164,11 @@ module Yast
     end
 
     # register all selected addons
+    # back returns directly to the extensions selection
     def register_addons
       return false if init_registration == :cancel
-      registration_ui.register_addons(@selected_addons, @known_reg_codes)
+      ret = registration_ui.register_addons(@selected_addons, @known_reg_codes)
+      (ret == :back) ? :extensions : ret
     end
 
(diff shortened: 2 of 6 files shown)


package/yast2-registration.changes
```diff
@@ -1,4 +1,12 @@
 -------------------------------------------------------------------
+Fri Feb 12 12:42:19 UTC 2016 - knut.anderssen@suse.com
+
+- Continue asking for registration codes also after failed
```
imobach: I would rephrase it. Maybe something like: `Ask again when an incorrect registration code is provided`.

teclator:
Going back ask again the last EULA code. @lslezak proposed going directly to the addons selection menu, so it's missing yet.

src/clients/inst_scc.rb
```diff
@@ -164,9 +164,11 @@ def get_available_addons
     end
 
     # register all selected addons
+    # back returns directly to the extensions selection
     def register_addons
       return false if init_registration == :cancel
-      registration_ui.register_addons(@selected_addons, @known_reg_codes)
+      ret = registration_ui.register_addons(@selected_addons, @known_reg_codes)
+      (ret == :back) ? :extensions : ret
```
lslezak: NP: I'd rather use

```ruby
ret = :extensions if ret == :back
ret
```

here as the `()`, `?` and `:` symbols make the line less readable...

lslezak:
LGTM, thanks for the fix! :+1:

src/lib/registration/registration_ui.rb
```diff
@@ -294,13 +294,16 @@ def register_selected_addons(selected_addons, known_reg_codes)
...
-        end
+            register_selected_addon(product, known_reg_codes[product.identifier])
+          end
 
         # remove from selected after successful registration
-        selected_addons.reject! { |selected| selected.identifier == product.identifier }
+        if registered
+          selected_addons.reject! { |selected| selected.identifier == product.identifier }
+        end
+        registered
```
mvidner: Before, the block returned the result of `reject!`. That seems inconsistent.

src/clients/inst_scc.rb
```diff
@@ -304,12 +306,14 @@ def start_workflow
...
-          next:  "update_autoyast_config"
+          abort:      :abort,
+          extensions: "select_addons",
+          next:       "update_autoyast_config"
         },
         "reregister_addons"      => {
-          abort: :abort,
-          next:  "check"
+          abort:      :abort,
+          extensions: "select_addons",
```
mvidner: Are you sure about this one? I have actually drawn the dialog transition graph for my feature, and this edge would continue to a different branch. Did you mean "select_addons_rereg" here?

src/lib/registration/registration_ui.rb
lslezak: Previously it returned `true` when the product was successfully removed from the selected list. That should happen in all cases, but it ignored the addon registration result. Now it honors the registration result which looks as a good change to me.

src/clients/inst_scc.rb
teclator: Thanks for that point!! sure i mean select_addons_rereg, i wii fix and test it manually

src/clients/inst_scc.rb
teclator: Ok, i will change it

src/lib/registration/registration_ui.rb
teclator: Exactly, before, it ever deleted the product from the selected addons, so now it register all with good codes and keep not register ones to be filled if you want to.

lslezak:
LGTM

PR #244, opened 2016-02-18
Support for using Jenkins CI

lslezak:
The main functionality is placed in the RPM macros, see https://github.com/yast/yast-devtools/pull/99.

### DIFF
--- a/package/yast2-registration.spec
+++ b/package/yast2-registration.spec
@@ -48,7 +48,7 @@ Requires:       yast2-update >= 3.1.19
 
 BuildRequires:  yast2 >= 3.1.26
 BuildRequires:  update-desktop-files
-BuildRequires:  yast2-devtools >= 3.1.6
+BuildRequires:  yast2-devtools >= 3.1.39
 BuildRequires:  rubygem(yast-rake) >= 0.2.5
 BuildRequires:  rubygem(rspec)
 BuildRequires:  rubygem(suse-connect) >= 0.2.22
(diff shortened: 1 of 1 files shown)


package/yast2-registration.spec
```diff
@@ -58,6 +58,13 @@ BuildRequires:  yast2-update >= 3.1.19
 
 BuildArch:      noarch
 
+# FIXME: it seems can we cannot move it to macros.yast, the yast-rake-ci is not
+# installed into the chroot, the build fails...
```
mvidner: I don't understand the comment.

mvidner: Ah, you mean moving this BuildRequires somewhere else?

What about declaring `%bcond_and_buildrequires_with_yast_run_ci_tests` in yast.macros, which expands to the lines below?

lslezak: Yes, I meant moving these lines to `macros.yast`.

Unfortunately it does not work, it seems that `osc` does not evaluate the spec file correctly when preparing the build root. It does not install the `rubygem-rake-ci` package into the build root. Later `rpmbuild` fails with missing BuildRequires error... :worried:

mvidner: I see. For a while I thought that it may be fixable, but then I looked into [the source](https://github.com/openSUSE/obs-build/blob/5f04f59b27bc1fbd1dfa908c5287657d306fe19e/Build/Rpm.pm#L191): a 400 line function in Perl. With a `goto`.

mvidner:
LGTM

PR #245, opened 2016-02-23
Do not perform SLP discovery when running specs.

mvidner:
Without this,
if there is a SLP server offering service:registration.suse on the local network,
and it is not blocked by iptables
the tests will loop endlessly.

See https://ci.suse.de/view/YaST/job/yast-registration-master/20/console , or turn off firewall on a machine in the Prague office and try `rake osc:build`, with and without this PR.


lslezak:
LGTM, thanks!

PR #246, opened 2016-03-03
Allow changing the build target via environment

lslezak:
Use `YAST_SUBMIT` variable to override the build target to [sle12sp1_public](https://github.com/yast/yast-rake/blob/3dc1fc49f9a17211a5c169070d1ae9872c69ee25/data/targets.yml#L16) at Jenkins builds.

See http://ci.opensuse.org/job/yast-registration-github-push/490//console.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -1,6 +1,7 @@
 require "yast/rake"
 
-Yast::Tasks.submit_to :sle12sp1
+yast_submit = ENV["YAST_SUBMIT"] || :sle12sp1
+Yast::Tasks.submit_to(yast_submit.to_sym)
 
 Yast::Tasks.configuration do |conf|
   # lets ignore license check for now


Rakefile
```diff
@@ -1,6 +1,7 @@
 require "yast/rake"
 
-Yast::Tasks.submit_to :sle12sp1
+yast_submit = ENV["YAST_SUBMIT"] || :sle12sp1
+Yast::Tasks.submit_to(yast_submit.to_sym)
```
mvidner: But YAST_SUBMIT already **is** used: https://github.com/yast/yast-rake/blob/2076875ee1c72dd9b4e16b9bb9e00a1c60c4e52b/lib/yast/rake.rb#L29
Except it has Factory as the default. But if you override the variable anyway, maybe you can simply remove this here?

lslezak: Well, removing the line would change the default target which would be a very bad side effect. So far you always use `rake osb:build` for running a testing build regardless the branch you currently use.

After removing the line you would need use `YAST_SUBMIT=sle12sp1 rake osc:build` in the *SLE-12-SP1* branch and the simpler `rake osc:build` for *master* otherwise you would build the SP1 code against Factory. That would be really confusing...

mvidner:
OK. LGTM.

PR #249, opened 2016-03-09
[WIP] Try to update at first run or after network configuration (FATE#319716)

teclator:
- This is related to yast/yast-installation#345


mvidner:
What is the status of this? Is this still needed?

teclator:
No it isn't. The first run update was moved to system analysis client and will be moved to a independent client, and try to update in case of not available network is under discussion, so i will close this PR by now.

PR #252, opened 2016-04-06
Unified shortcut 'alt-w' for network setup button (bsc#974216)

teclator:
registrationnetshortcut


lslezak:
NP: Maybe we should share the label in the `Label` module, just like we define [AddButton](https://github.com/yast/yast-yast2/blob/master/library/general/src/modules/Label.rb#L44) or others...

Anyway, LGTM, we can move it to the `Label` module later next time we touch it...

PR #253, opened 2016-04-06
forcing refreshing

schubi2:


mvidner:
The dependency is https://github.com/yast/yast-pkg-bindings/pull/58

package/yast2-registration.changes
```diff
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Wed Apr  6 16:10:58 CEST 2016 - schubi@suse.de
+
+- Force refreshing while adding a service (bnc#967828).
```
mvidner: OK, but why? Reading the bug, I conclude "otherwise registration can fail". Well, it may be obvious, but it is still more clear to state it explicitly.

mvidner:
The Jenkins build is failing because the dependency is missing, but it will fail even with the proper dependency because of this: https://github.com/yast/yast-registration/search?utf8=%E2%9C%93&q=ServiceRefresh

lslezak:
I just agree what @mvidner wrote: improve the changes and fix the test as well.

Thanks!

lslezak:
LGTM, thanks! :+1:

PR #254, opened 2016-04-08
Bug 967828 master

schubi2:
Backport from sles12-sp1


lslezak:
LGTM, looks like some Travis issue, I have updated `yast2-update` package, let's see if that helps...

Anyway, we can ignore Travis as we already have Jenkins support which passes. :wink:

mvidner:
LGTM

PR #255, opened 2016-05-11
Select the default product patterns (FATE#320199)

lslezak:
[FATE#320199](https://fate.suse.com/320199) - Select the default product patterns, this part is for the registered modules and extensions in an installed system.

- 3.1.172

### Testing

The code has been tested in a patched SP1 installation as SP2 builds cannot be registered yet.

Here is a snippet from `y2log` showing that the code was called as expected after registering the *Web and Scripting* module:

```
packager/product_patterns.rb:54 Found selected products: ["sle-module-web-scripting"]
packager/product_patterns.rb:140 Found default patterns: []
packager/product_patterns.rb:57 Default patterns for the selected products: []
registration/sw_mgmt.rb:469 Selecting the default product patterns: []
```

*Note: No product was selected as the new `defaultpattern()` tags are not used in SP1 products, this just proves that the pattern selection code was executed.*

### DIFF
--- a/package/yast2-registration.spec
+++ b/package/yast2-registration.spec
@@ -17,7 +17,7 @@
 
 
 Name:           yast2-registration
-Version:        3.1.171
+Version:        3.1.172
 Release:        0
 
 BuildRoot:      %{_tmppath}/%{name}-%{version}-build
@@ -43,7 +43,8 @@ Requires:       SUSEConnect
 
 Requires:       yast2-slp >= 3.1.2
 Requires:       yast2-add-on >= 3.1.8
-Requires:       yast2-packager >= 3.1.26
+# packager/product_patterns.rb
+Requires:       yast2-packager >= 3.1.95
 Requires:       yast2-update >= 3.1.36
 
 BuildRequires:  yast2 >= 3.1.26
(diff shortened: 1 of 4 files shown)


package/yast2-registration.spec
```diff
@@ -43,7 +43,8 @@ Requires:       SUSEConnect
 
 Requires:       yast2-slp >= 3.1.2
 Requires:       yast2-add-on >= 3.1.8
-Requires:       yast2-packager >= 3.1.26
+# packager/product_patterns.rb
+Requires:       yast2-packager >= 3.1.95
```
teclator: Why not use >= 3.1.97 directly?, should not affect but as you are changing it.

teclator:
LGTM

package/yast2-registration.spec
lslezak: 3.1.95 introduced the `packager/product_patterns.rb` file, 3.1.97 just changed the place where the new code was used in the packager itself. The API haven't changed since 3.1.95, so it's the correct version.

(Anyway, for the end users it's irrelevant, both 3.1.95 and 3.1.97 were not released to public, they will see 3.1.98+ :wink:)

PR #256, opened 2016-05-24
New registration UI

imobach:
[Six sprints ago](https://lizards.opensuse.org/2016/02/03/highlights-of-development-sprint-14/), the "Local User" screen got [some love](https://github.com/yast/yast-users/pull/84) and the UI was greatly improved. During this sprint, and according to [bug #974626](https://bugzilla.suse.com/show_bug.cgi?id=974626), we have improved the registration UI to make it consistent with the "Local User" screen.

## The old (and confusing) dialog

old-registration-ui

The old interface was pretty confusing. To make it clear, you have three options:

1. Register the system using scc.suse.com by introducing an e-mail and a registration code.
2. Register the system using a local SMT server (no e-mail or registration code are used).
3. Skip the registration step.

Options 1 and 2 are mutually exclusive but, if you look at this interface, that fact is not clear.

On the other hand, we wanted this dialog to be consistent with the new "Local User" one.

## The new dialog

new-registration-ui-qt

The new dialog is pretty clear: you have three mutually exclusive options and that's all. And what about the ncurses interface? Well, you can see how it looks like below (it fits in a 80x25 screen).

new-registration-ui-ncurses

new-registration-ui-skip-ncurses

## Code refactoring

As part of this change, we made some refactor and clean-up some old code. And, of course, we also improved test coverage for this dialog.


src/lib/registration/ui/base_system_registration_dialog.rb
```diff
@@ -52,10 +64,103 @@ def run
...
 
-        handle_dialog
+      # Handle selecting the 'Skip' option
+      #
+      # Set the dialog's action to :skip_registration
+      def skip_registration_handler
+        self.action = :skip_registration
+      end
+
+      # Handle selection the 'Register System via SCC.SUSE.COM' option
```
imobach: I'm not sure about upcasing `SCC.SUSE.COM`. I need to check it with Ken.

src/lib/registration/ui/base_system_registration_dialog.rb
```diff
@@ -78,23 +178,123 @@ def content
...
+      end
+
+      def register_scc_option
+        VBox(
+          Left(
+            RadioButton(
+              Id(:register_scc),
+              Opt(:notify),
+              # TRANSLATORS: radio button
+              _("Register System via SCC.SUSE.COM"),
```
imobach: I'm not sure about upcasing `SCC.SUSE.COM`. I need to check it with Ken.

lslezak:
I'd prefer lowercase SCC host name, the URL or host names are usually (maybe rather always) written in lower case.

src/lib/registration/ui/base_system_registration_dialog.rb
lslezak: I'd prefer lowercase SCC host name, the URL or host names are usually (maybe rather always) written in lower case.

lslezak:
So far I checked the screenshots:

- I'd move the "*If you skip product registration now, remember to register after installation has completed.*" text below the radio button. The text is related to skipping the registration so it should be located near the *Skip Registration* radio button.

- The text mode screenshot looks a bit crowded at the bottom IMHO. I'd remove the spacing at the top or between the radio buttons.

Ideally you could obtain the text mode screen size via `UI.GetDisplayInfo` call which in text mode returns keys `Height` and `Width` with screen dimensions in characters. Then add the extra vertical spacing only if `Height` > 28 (or something like this) or if a graphical mode is used.

imobach:
@lslezak I agree about removing the spacing (thanks for the trick). On the other hand, I'm not so sure about moving the skip text.

kwwii:
@lslezak I agree with your comment. Ideally, we should show a pop-up with the text when the user selects the "skip registration" radio button, or if that's not possible when the user has selected that and then clicks on "next"

imobach:
@kwwii Showing a pop-up is possible, so I'll change that. As a side effect, that would also save some space.

imobach:
@lslezak @kwwii Currently, we show this message when the user click on "next" (and the "skip" option is selected):

```
If you do not register your system we will not be able
to grant you access to the update repositories.

You can register after the installation or visit our
Customer Center for online registration.

Really skip the registration now?
```

Maybe we could show just the warning (without asking for confirmation) when the user `selects` the option. Thoughts?

lslezak:
And what about the `https://scc.suse.com` default for the SMT server address? This default does not make much sense IMHO. Selecting the SMT option and then using (leaving) the SCC server URL is equivalent to selecting the first option, it sounds a bit weird to me.

IMHO we should either leave the field empty by default, or use some example name there, e.g. `https://smt.example.com`. In the latter case users will see what is expected to enter there and if they are lucky and use the `smt` host name for their SMT server then they will need to edit just the domain name at the end and save some typing. (BTW we use https://smt.suse.cz and https://smt.suse.de internally :wink: )

lslezak:
> Maybe we could show just the warning (without asking for confirmation) when the user selects the option. Thoughts?

Um, displaying the popup just after *selecting* the option would require some changes to the text and type of the popup. The final question "*Really skip the registration now?*" would look strange. More over what would do if user pressed [No]? Change the selected radio button?

If we want to display the popup just after selecting the radio button the last question should be removed and it should be changed to a simple popup with single [Ok] button only (instead of the Yes/No).

@kwii?

src/lib/registration/ui/base_system_registration_dialog.rb
```diff
@@ -419,9 +407,12 @@ def set_registration_options
         options = Storage::InstallationOptions.instance
         case action
         when :register_scc
-          options.email = Yast::UI.QueryWidget(:email, :Value)
-          options.reg_code = Yast::UI.QueryWidget(:reg_code, :Value)
+          options.email      = Yast::UI.QueryWidget(:email, :Value)
+          options.reg_code   = Yast::UI.QueryWidget(:reg_code, :Value)
+          options.custom_url = nil
         when :register_local
+          options.email      = "" # Use an empty string like InstallationOptions constructor
```
lslezak: NP: I'd prefer adding the comment above the code, trailing comments usually make the lines quite long...

src/lib/registration/ui/base_system_registration_dialog.rb
```diff
@@ -328,41 +341,16 @@ def help_text
...
-          # label text describing the registration (2/2),
-          # not displayed in installed system
-          # use \n to split to more lines if needed (use max. 76 chars/line)
-          info += _("If you skip product registration now, remember to register after\n" \
-              "installation has completed.")
-        end
-
-        info
+            "Customer Center for online registration.")
+        Yast::Popup.Message(warning)
```
lslezak: NP: A small inconsistency, I'd use `warning` variable and `Popup.Warning` method or `message` variable and `Popup.Message` method. With this code it's not obvious what is the intention. Is it a warning or an informative message?

lslezak:
I have checked the added commits, no new issue found. :smile:

imobach:
@lslezak Thanks! Now I'm looking for a LGTM :) On the other hand, I'll report the issue I found with certificates (which is not caused by yast2-registration).

lslezak:
LGTM, thanks! :+1:

imobach:
I've updated the PBI to take the regurl parameter into account. Now if such a parameter is specified, the second option (local SMT server) is selected using the regurl value.

lslezak:
Thanks, LGTM!

lslezak:
:tada:

digitaltom:
Nice improvement! 
Just one remark and one question: 
The email address is an optional field, maybe you want to mark it as such. 
The note "or visit our Customer Center for online registration", what does that actually mean? There is no way to register a system online (via browser). 
Maybe it would be better to write "You can still register your system after installation by using YaST or SUSEConnect."

imobach:
@digitaltom Thanks for your feedback! The message looks wrong, so I've created [bsc#982560](https://bugzilla.suse.com/show_bug.cgi?id=982560) to discuss the best option with @kwwii.

PR #257, opened 2016-06-03
regurl overwrites default URL registration

imobach:
It should fix the issue with using a SCC proxy.


lslezak:
LGTM, thanks for a very quick fix! :+1:

imobach:
Thanks for review!

PR #258, opened 2016-06-14
Preselect the Toolchain module on ARM64

lslezak:
## Details

- See [FATE320679](https://fate.suse.com/320679)
- There are also few test suite improvelemnts
- The workaround is limited to SP2, in SP3 we will need a proper solution

## Testing

- Added unit tests
- Tested in a patched SLES installation (the latest build)
- I have manually adjusted the patch with simple `s/aarch64/x86_64/g` so it could be tested on x86_64 without need for special ARM64 hardware:

# Screencast
*(Adjusted and tested on x86_64 to be easier to test, in reality the module is not pre-selected on x86_64.)*

preselect_toolchain6

### DIFF
--- a/src/lib/registration/ui/addon_selection_base_dialog.rb
+++ b/src/lib/registration/ui/addon_selection_base_dialog.rb
@@ -268,6 +268,7 @@ module Registration
 
       # workaround for FATE#320679 - preselect the Toolchain module on ARM
       # in SLES12-SP2
+      # FIXME: remove this hack in SLES12-SP3, use a proper solution instead
       def aarch64_workaround
         # SLES12-SP2 base?
         product = SwMgmt.base_product_to_register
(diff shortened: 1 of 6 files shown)


src/lib/registration/ui/addon_selection_base_dialog.rb
```diff
@@ -260,6 +265,25 @@ def generic_help_text
           _("<p>If you want to remove any extension or module you need to log"\
               "into the SUSE Customer Center and remove them manually there.</p>")
       end
+
+      # workaround for FATE#320679 - preselect the Toolchain module on ARM
+      # in SLES12-SP2
```
kobliha: Some TODO or FIXME or XXX would be nice (we want to remove it after SLE 12 SP2)

imobach:
LGTM

PR #259, opened 2016-06-17
Offer a list of found local SMT servers

mvidner:
https://bugzilla.suse.com/show_bug.cgi?id=981633
(internal tracking: https://trello.com/c/vwuEhHKZ)

Before, the widget was a single text field and a little helpful `smt.example.com` was always shown (no matter what your actual domain was).
local-smt-server-1-before

Now, if your local [SMT](https://www.suse.com/documentation/smt11/) servers are advertised via [SLP](https://en.wikipedia.org/wiki/Service_Location_Protocol) they will be offered as choices . (Here `acme.com` stands for your domain)
local-smt-server-2-after

### DIFF
--- a/src/lib/registration/ui/base_system_registration_dialog.rb
+++ b/src/lib/registration/ui/base_system_registration_dialog.rb
@@ -279,8 +279,15 @@ module Registration
       #
       # @return [Yast::Term] UI terms
       def register_local_option
-        # If not special URL is used, show an example one.
-        url = using_default_url? ? EXAMPLE_SMT_URL : reg_options[:custom_url]
+        # If not special URL is used, probe with SLP
+        if using_default_url?
+          services = UrlHelpers.slp_discovery_feedback
+          urls = services.map { |svc| UrlHelpers.service_url(svc.slp_url) }
+          urls = [EXAMPLE_SMT_URL] if urls.empty?
+        else
+          urls = [reg_options[:custom_url]]
+        end
+
         VBox(
           Left(
             RadioButton(
@@ -296,8 +303,10 @@ module Registration
             HBox(
               HSpacing(5),
               VBox(
-                MinWidth(REG_CODE_WIDTH, InputField(Id(:custom_url),
-                  _("&Local Registration Server URL"), url))
+                MinWidth(REG_CODE_WIDTH,
+                  ComboBox(Id(:custom_url), Opt(:editable),
+                    _("&Local Registration Server URL"), urls)
+                )
               )
             )
           ),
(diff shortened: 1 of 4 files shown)


mvidner:
In the Trello card, one of the tasks is explicitly:
-  List of SLP servers collected in advance, available also for other purposes (self-update)

I am not sure what that means. I simply called a library function which did the SLP lookup.

src/lib/registration/ui/base_system_registration_dialog.rb
```diff
@@ -272,15 +272,18 @@ def register_scc_option
         )
       end
 
-      # Example URL to be used in the :register_local UI
-      EXAMPLE_SMT_URL = "https://smt.example.com"
```
lslezak: Well, the idea for pre-filling the `https://smt.example.com` default is that:

- Users will clearly see what input is expected
- In many cases users will just need to update the domain name at the end of the URL and save some typing (locally we use `https://smt.suse.cz` and `https://smt.suse.de` :wink: )

But this does not work well with the editable combo box...

Maybe use the `https://smt.example.com` default only when the SLP scan does not find any server? And if something is found display just the found servers without the default which could confuse the users... :question:

lslezak:
That means the SLP scan should be started earlier in the workflow and the result could be also used for the self-update feature. The reason is probably to avoid repeated SLP scans.

But this caching/sharing feature would make it more complicated. Maybe we should discuss this with the PO, how much this feature is required... @kobliha :question: 

BTW it seems that the SLP tool/library has a maximum time for scanning set to 5 seconds (see `/etc/slp.conf`), but in my tests
```
time slptool findsrvs service:registration.suse
```
always prints 2.506s real time.

Anyway, I do not see any problem with repeated scans, it takes just few seconds. And there is set a reasonable maximum limit so it's guaranteed it cannot get stuck waiting for a response forever...

lslezak:
In theory we could run the scan in background, we could start it just after configuring the network and by the time user gets into the registration step the SLP scan would be finished.

But that would require a different implementation, currently we use the `.slp` agent which cannot be run in   background, we would need to use the `slptool` command line tool...

mvidner:
I see. In my tests, the lookup is even faster (1.5s) and we already do show a pop-up with "Looking up local registration servers..." while the lookup is in progress. If the library timeout for SLP lookup is 5s then I think the time savings is not worth the code complication and I would like to scratch that.

lslezak:
> If the library timeout for SLP lookup is 5s then I think the time savings is not worth the code complication and I would like to scratch that.

Fully agreed, the scanning in advance would make the code much more complicated and the time savings would be just few seconds at most. Not worth doing it IMHO...

lslezak:
BTW the code change LGTM

PR #260, opened 2016-07-15
Rubocop update

lslezak:
Notes:

- *Lint/ShadowedException* has been disabled as it cannot handle [this complex case](https://github.com/yast/yast-registration/blob/cde3fb50db6d2b70cac7b9a203151b49e61b9bf8/src/lib/registration/connect_helpers.rb#L70) correctly. I'll investigate it later.
- Jenkins fails because the `rubygem-rubocop.rpm` has not been updated in YaST:Head yet (I'll do it, I have added a task to the Trello card). Let's ignore this failure for now, Travis passes.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -29,6 +29,12 @@ Metrics/MethodLength:
 Metrics/PerceivedComplexity:
   Max: 25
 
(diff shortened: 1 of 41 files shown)


jreidinger:
LGTM

PR #261, opened 2016-07-20
Make many extensions fit on the screen with a scrollable widget

mvidner:
References:
Scrum: https://trello.com/c/L7i4X9kv
Bug: [bsc#967387](https://bugzilla.suse.com/show_bug.cgi?id=967387)
Also includes [FATE#319909](https://fate.suse.com/319909) - Do not show beta versions of products in YaST extension listing

### DIFF
--- a/src/lib/registration/ui/addon_selection_base_dialog.rb
+++ b/src/lib/registration/ui/addon_selection_base_dialog.rb
@@ -218,13 +218,8 @@ module Registration
         addon = @addons.find { |a| addon_widget_id(a) == id }
         return unless addon
 
+        addon.toggle_selected
         show_addon_details(addon)
-        new_state = !addon.selected?
-        if new_state
-          addon.selected
-        else
-          addon.unselected
-        end
         reactivate_dependencies
       end
 
--- a/src/lib/registration/ui/addon_selection_base_dialog.rb
+++ b/src/lib/registration/ui/addon_selection_base_dialog.rb
@@ -147,69 +147,6 @@ module Registration
         VWeight(75, MinHeight(12, content))
       end
 
-      # display the addon checkboxes in two columns
-      # @param col1 [Array<Addon>] the addons displayed in the first column
-      # @param col2 [Array<Addon>] the addons displayed in the second column
-      # @return [Yast::Term] the addon cheboxes
-      def two_column_layout(col1, col2)
-        box2 = addon_selection_items(col1)
-        box2.params << VStretch() # just UI tweak
-
-        HBox(
-          addon_selection_items(col2),
-          HSpacing(1),
-          box2
-        )
-      end
-
-      # create a single UI column with addon checkboxes
-      # @return [Yast::Term] addon column
-      def addon_selection_items(addons)
-        box = VBox()
-
-        # whether to add extra spacing in the UI
-        add_extra_spacing = if Yast::UI.TextMode
-          addons.size < 5
-        else
-          true
-        end
-
-        addons.each do |addon|
-          box.params.concat(addon_checkbox(addon, add_extra_spacing))
-        end
-
-        box
-      end
-
-      # create spacing around the addon checkbox so the layout looks better
-      # @param addon [Registration::Addon]
-      # @param extra_spacing [Boolean] add extra spacing (indicates enough space in UI)
-      # @return [Array<Yast::Term>] Return array with one or two elements for VBox
-      def addon_checkbox(addon, extra_spacing)
-        checkbox = Left(addon_checkbox_element(addon))
-
-        # usability help. If addon depends on something, then we get it
-        # immediatelly after parent, so indent it slightly, so it is easier visible
-        checkbox = HBox(HSpacing(2.5), checkbox) if addon.depends_on
-
-        res = [checkbox]
-        # add extra spacing when there are just few addons, in GUI always
-        res << VSpacing(0.7) if extra_spacing
-
-        res
-      end
-
-      # create the UI checkbox element for the addon
-      # @param addon [Registration::Addon] the addon
-      # @return [Yast::Term] checkbox term
-      def addon_checkbox_element(addon)
-        # checkbox label for an unavailable extension
-        # (%s is an extension name)
-        label = addon.available? ? addon.label : (_("%s (not available)") % addon.label)
-
-        CheckBox(Id(addon_widget_id(addon)), Opt(:notify), label, addon_selected?(addon))
-      end
-
       # the main event loop - handle the user in put in the dialog
       # @return [Symbol] the user input
       def handle_dialog
--- a/src/lib/registration/ui/addon_selection_base_dialog.rb
+++ b/src/lib/registration/ui/addon_selection_base_dialog.rb
@@ -98,10 +98,10 @@ module Registration
           # (%s is an extension name)
           label = addon.available? ? addon.label : (_("%s (not available)") % addon.label)
 
-          rt_cb(id: addon_widget_id(addon),
-                label: label,
+          rt_cb(id:       addon_widget_id(addon),
+                label:    label,
                 selected: addon_selected?(addon),
-                enabled: addon.available?,
+                enabled:  addon.available?,
                 indented: addon.depends_on)
         end
         items.join("\n")
@@ -111,11 +111,11 @@ module Registration
       # https://gist.github.com/lslezak/5ed82fe19337bef4807b
 
       # FIXME: adapt to installation theme
-      IMG_ON = "/usr/share/YaST2/theme/current/wizard/checkbox-on.png"
-      IMG_OFF = "/usr/share/YaST2/theme/current/wizard/checkbox-off.png"
+      IMG_ON = "/usr/share/YaST2/theme/current/wizard/checkbox-on.png".freeze
+      IMG_OFF = "/usr/share/YaST2/theme/current/wizard/checkbox-off.png".freeze
 
       def rt_cb(id:, label:, selected:, enabled:,
-                indented: false, text_mode: Yast::UI.TextMode)
+        indented: false, text_mode: Yast::UI.TextMode)
         selected = false unless enabled
         if text_mode
           indent = "&nbsp;" * (indented ? 5 : 1)
@@ -190,7 +190,7 @@ module Registration
         return unless addon
 
         show_addon_details(addon)
-        new_state = ! addon.selected?
+        new_state = !addon.selected?
         if new_state
           addon.selected
         else
--- a/src/lib/registration/ui/addon_selection_base_dialog.rb
+++ b/src/lib/registration/ui/addon_selection_base_dialog.rb
@@ -29,17 +29,19 @@ module Registration
       #   communication with SCC
       def initialize(registration)
         textdomain "registration"
-        @addons = Addon.find_all(registration)
+        @all_addons = Addon.find_all(registration)
 
         # sort the addons
-        @addons.sort!(&::Registration::ADDON_SORTER)
+        @all_addons.sort!(&::Registration::ADDON_SORTER)
+
+        filter_beta_releases(true)
 
         @old_selection = Addon.selected.dup
 
         # activate a workaround on ARM (FATE#320679)
         aarch64_workaround if Arch.aarch64
 
-        log.info "Available addons: #{@addons}"
+        log.info "Available addons: #{@all_addons}"
       end
 
       # reimplement this in a subclass
@@ -56,6 +58,12 @@ module Registration
       # @return [String] widget id
       def addon_widget_id(addon)
         "#{addon.identifier}-#{addon.version}-#{addon.arch}"
+      end
+
+      # Enables or disables beta addons filtering
+      # @param [Boolean] enable true for filtering beta releases
+      def filter_beta_releases(enable)
+        @addons = enable ? @all_addons.reject(&:beta_release?) : @all_addons
       end
 
     private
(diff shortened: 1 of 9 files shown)


src/lib/registration/ui/addon_selection_base_dialog.rb
```diff
@@ -93,16 +93,51 @@ def details_widget
...
+
+          rt_cb(id: addon_widget_id(addon),
+                label: label,
+                selected: addon_selected?(addon),
+                enabled: addon.available?,
+                indented: addon.depends_on)
+        end
+      end
+
+      IMG_ON = "/usr/share/YaST2/theme/current/wizard/checkbox-on.png"
```
lslezak: Um, I just realized that we need to handle a different UI theming in installation. So in `Stage.initial` we need to use https://github.com/yast/yast-theme/blob/master/SLE/wizard/inst_checkbox-on.png and https://github.com/yast/yast-theme/blob/master/SLE/wizard/inst_checkbox-off.png images.

#### Installation



#### Installed system

src/lib/registration/ui/addon_selection_base_dialog.rb
```diff
@@ -93,18 +93,56 @@ def details_widget
...
+      end
+
+      # FIXME acknowledge Lada's code
+      # https://gist.github.com/lslezak/5ed82fe19337bef4807b
+
+      IMG_ON = "/usr/share/YaST2/theme/current/wizard/checkbox-on.png"
+      IMG_OFF = "/usr/share/YaST2/theme/current/wizard/checkbox-off.png"
+
+      def rt_cb(id:, label:, selected:, enabled:,
+                indented: false, text_mode: Yast::UI.TextMode)
```
jreidinger: I really do not like this API:

1. too much parameters
2. why not passing addon object directly?
3. why text mode parameter is needed at all?
4. label should be also constructed here

src/lib/registration/ui/addon_selection_base_dialog.rb
```diff
@@ -93,18 +93,56 @@ def details_widget
...
+        selected = false unless enabled
+        if text_mode
+          indent = "&nbsp;" * (indented ? 5 : 1)
+          check = selected ? "[x]" : "[ ]"
+          widget = "#{check} #{label}"
+          enabled_widget = enabled ? "<a href=\"#{id}\">#{widget}</a>" : widget
```
jreidinger: I do not see in text mode any visual indication that  addon is disabled except a link, which is not enough from my POV. I think that maybe have check something like "---" will be better?

src/lib/registration/ui/addon_selection_base_dialog.rb
```diff
@@ -93,18 +93,56 @@ def details_widget
...
       # the UI uses two column layout
       # @return [Yast::Term] the main UI dialog term
       def addons_box
-        lines = Yast::UI.TextMode ? 9 : 14
-        if @addons.size <= lines
-          content = addon_selection_items(@addons)
-        else
-          content = two_column_layout(@addons[lines..(2 * lines - 1)], @addons[0..(lines - 1)])
-        end
+        content = RichText(Id(:items), richtext_checkboxes(@addons))
```
jreidinger: when you modify UI, can you please also add to top link to enable/disable beta add-ons filtering? I think it fit there and to avoid conflicts I do not add it to my branch.

src/lib/registration/ui/addon_selection_base_dialog.rb
```diff
@@ -211,7 +249,8 @@ def handle_addon_selection(id)
         return unless addon
 
         show_addon_details(addon)
-        if Yast::UI.QueryWidget(Id(addon_widget_id(addon)), :Value)
+        new_state = ! addon.selected?
+        if new_state
```
jreidinger: why not simple `addon.selected? ? addon.unselected : addon.selected` or adding it to addon class with method like toggle_selected?

src/lib/registration/ui/addon_selection_base_dialog.rb
```diff
@@ -229,9 +268,10 @@ def show_addon_details(addon)
 
       # update the enabled/disabled status in UI for dependent addons
       def reactivate_dependencies
-        @addons.each do |addon|
-          Yast::UI.ChangeWidget(Id(addon_widget_id(addon)), :Enabled, addon.selectable?)
-        end
+        Yast::UI.ChangeWidget(Id(:items), :Value, richtext_checkboxes(@addons))
+#        @addons.each do |addon|
+#          Yast::UI.ChangeWidget(Id(addon_widget_id(addon)), :Enabled, addon.selectable?)
+#        end
```
jreidinger: dead code

src/lib/registration/ui/addon_selection_registration_dialog.rb
```diff
@@ -59,9 +59,10 @@ def addon_selected?(addon)
 
       # update the enabled/disabled status in UI for dependent addons
       def reactivate_dependencies
-        @addons.each do |addon|
-          Yast::UI.ChangeWidget(Id(addon_widget_id(addon)), :Enabled, addon.selectable?)
-        end
+        Yast::UI.ChangeWidget(Id(:items), :Value, richtext_checkboxes(@addons))
+#        @addons.each do |addon|
+#          Yast::UI.ChangeWidget(Id(addon_widget_id(addon)), :Enabled, addon.selectable?)
+#        end
```
jreidinger: dead code

mvidner:
This now includes #262.

src/lib/registration/ui/addon_selection_base_dialog.rb
```diff
@@ -93,86 +101,83 @@ def details_widget
...
+        items = addons.map do |addon|
+          # checkbox label for an unavailable extension
+          # (%s is an extension name)
+          label = addon.available? ? addon.label : (_("%s (not available)") % addon.label)
+
+          richtext_checkbox(id:       addon_widget_id(addon),
+                            label:    label,
+                            selected: addon_selected?(addon),
+                            enabled:  addon.available?,
+                            indented: addon.depends_on)
```
jreidinger: I still do not get why it is better to send 5 parameters instead one `addon`?

src/lib/registration/ui/addon_selection_base_dialog.rb
```diff
@@ -93,86 +101,83 @@ def details_widget
...
-        end
+          indent = "&nbsp;" * (indented ? 7 : 1)
 
-        addons.each do |addon|
-          box.params.concat(addon_checkbox(addon, add_extra_spacing))
-        end
+          installation = !Yast::Mode.normal
+          image = (installation ? "inst:" : "normal:") +
+            (selected ? "on:" : "off:") + (enabled ? "enabled" : "disabled")
+          color = installation ? "white" : "black"
```
jreidinger: cannot we get somehow color from theme? this is quite fragile when style will be changed

src/lib/registration/ui/addon_selection_base_dialog.rb
```diff
@@ -211,7 +219,8 @@ def handle_addon_selection(id)
         return unless addon
 
         show_addon_details(addon)
-        if Yast::UI.QueryWidget(Id(addon_widget_id(addon)), :Value)
+        new_state = !addon.selected?
+        if new_state
```
jreidinger: still my comment regrding simplifying to `addon.selected? ? addon.unselected : addon.selected` is valid

src/lib/registration/ui/addon_selection_base_dialog.rb
mvidner: Yes, it is worse in that we send many more parameters. But the better thing is that the method does not have to care about the argument being an `Addon`.

src/lib/registration/ui/addon_selection_base_dialog.rb
mvidner: Yes, it is fragile. I have discussed this with Ken. We thought of using qss classes, but since this is a RichText, it uses yet another stylesheet, and it will not work out.

src/lib/registration/ui/addon_selection_base_dialog.rb
mvidner: I chose an explicit negation to express my intent of toggling the current state when the widget is clicked. I will use `toggle_selected` like you suggested earlier.

src/lib/registration/ui/addon_selection_base_dialog.rb
jreidinger: well, if you want to reuse that method later as independent method then it make sense. But if you want to keep it here, then it make more sense to me to rename it to addon_checkbox with single parameter. In fact those two can co-existists, with the second one being on top of the first one.

src/lib/registration/ui/addon_selection_base_dialog.rb
```diff
@@ -93,86 +101,84 @@ def details_widget
...
-        res
+          check = "<img src='#{IMAGE_DIR}/#{IMAGES[image]}'></img>"
+          widget = "#{check} #{label}"
+          enabled_widget = if enabled
+            "<a href='#{id}' style='text-decoration:none; color:#{color}'>#{widget}</a>"
+          else
+            "<span style='color:grey'>#{widget}</span>"
+          end
+          "<p>#{indent}#{enabled_widget}</p>"
+        end
```
jreidinger: my sandi metz part of brain see that it can be nice example for objects, like factory for checkbox representation, deciding if use text one or graphical one, with its elements and so on, but it is out of scope this pr.

src/lib/registration/ui/addon_selection_reregistration_dialog.rb
```diff
@@ -44,6 +44,8 @@ def run
           true
         )
 
+        reactivate_dependencies
+
```
jreidinger: why it is needed?

src/lib/registration/ui/addon_selection_reregistration_dialog.rb
```diff
@@ -62,6 +64,8 @@ def addon_selected?(addon)
       # empty implementation, allow reregistration of a dependant addon
       # without reregistering its parent
       def reactivate_dependencies
+        # FIXME: how to test this?
```
jreidinger: I do not get this fixme, what needed to test?
also comment above is no longer correct

mvidner: The fixme is to remind us that it is not covered by unit tests and until recently I did not know how to triigger it in the UI, but now I know.

mvidner: But the comment points out a naming problem: the method was  actually a blend of a generic hook which called a specific action (reactivating the dependencies), but now it needs 2 actions, adding a refresh of the richtext widget. I'll fix that.

jreidinger:
LGTM

digitaltom:
Nice improvement! As a sidenote, I don't think it's needed to display the architecture for each extension/module. It will always be the same architecture as the base system, and removing it would make the listing look nicer.

lslezak:
@digitaltom: The architecture is included in the `friendly_name` in the SCC response:

```
#<OpenStruct id=1153, name="Web and Scripting Module", identifier=
"sle-module-web-scripting", former_identifier="sle-module-web-scripting", version="12",
release_type=nil, arch="x86_64", friendly_name="Web and Scripting Module 12 x86_64" ...
```

`friendly_name` is preferred over `name` by YaST: https://github.com/yast/yast-registration/blob/e8c456677bff39646e304fafc0f60ef24b3e9d9d/src/lib/registration/addon.rb#L169

src/lib/registration/ui/addon_selection_base_dialog.rb
```diff
@@ -76,12 +86,12 @@ def addon_selected?(_addon)
       # @return [Yast::Term] the main UI dialog term
       def content
         VBox(
-          VStretch(),
           Left(Heading(heading)),
+          Left(CheckBox(Id(:filter_beta), Opt(:notify),
```
lslezak: Um, what about hiding the `Filter Out Beta Versions` checkbox when there is no Beta add-on available? To me it looks a bit confusing when you change the check box state and the list does not change... 

Just tried the code with SLE12-GA (using the `FAKE_BASE_PRODUCT=1` hack) and I found the behavior a bit strange, the first idea was that the filtering was broken...

lslezak: Moreover the [relevant SCC Trello card](https://trello.com/c/Dj8XvuhA/3323-2-api-of-connect-should-have-information-about-release-state-of-product) says that there might be more states in the future (`alpha`, `pre-alpha`,...). So maybe we should invert the logic and rename the filter to something like `Show only released versions` or `Show only supported versions`...

mvidner: @lslezak I see. Your points make sense but I think they are relatively minor at this point so I will not rush to fix them.

jsrain: For the alpha / pre-alpha, I don't think that we need this for SP2; hiding the beta check-box if there are none available is something that we should IMO handle before GM (e.g. next sprint), the way should be, however, discussed with UX experts (customers could ask "where is the check box?")

kwwii: I think we should show this all the time so that the user learns and knows where it is. Instead of "Filter out Beta Products" I would change the wording/logic to "Show Beta Products (if available)" and no pre-select the checkbox. A user can then show the beta products, if desired, by selecting the checkbox.

PR #262, opened 2016-07-20
[no merge] UI change for many extensions

jreidinger:


src/lib/registration/ui/addon_selection_base_dialog.rb
```diff
@@ -29,17 +29,19 @@ class AddonSelectionBaseDialog
...
       def initialize(registration)
         textdomain "registration"
-        @addons = Addon.find_all(registration)
+        @all_addons = Addon.find_all(registration)
 
         # sort the addons
-        @addons.sort!(&::Registration::ADDON_SORTER)
+        @all_addons.sort!(&::Registration::ADDON_SORTER)
+
+        @addons = @all_addons.reject(&:beta_release?)
```
lslezak: DRY: I'd prefer `filter_beta_releases(true)` here...

jreidinger: you are right, will change it

test/addon_spec.rb
```diff
@@ -146,6 +146,20 @@
     end
   end
 
+  describe "#beta_release?" do
+    it "returns if addon is beta release" do
```
lslezak: NP: I'd split it to two tests like `it returns true if addon is beta release` and `it returns false if addon is not beta release`.

PR #263, opened 2016-07-21
Add support to integrate SCC/SMT and installer updates

imobachgs:
This PR includes changes needed for https://github.com/yast/yast-installation/pull/401:

* Extract service selection dialog into a separate class (inheriting from [ServiceSelectionDialog](https://github.com/yast/yast-slp/pull/17)).
* Add a new method to get installer updates in `Registration::Registration` class.


src/clients/discover_registration_services.rb
```diff
@@ -28,91 +25,13 @@ def main
...
-    end
-
-    def service_url(selected)
-      selected_service = services[selected.to_i]
-      log.info "Selected service #{selected_service.inspect}"
-
-      url = ::Registration::UrlHelpers.service_url(selected_service.slp_url)
+    def select_registration_service
+      service = ::Registration::UI::ServiceSelectionDialog.run(services)
+      return nil unless service.respond_to?(:slp_url)
```
jreidinger: looks a bit strange. I expect that here you at max handle abort, but some logic regarding url I will expect in service_url helper if needed.

jreidinger: ok, seeing logic below I expect here to be full handling...e.g. what happen if user select slp and then go back and select scc? it should be set back, not?

src/lib/registration/ui/service_selection_dialog.rb
```diff
@@ -0,0 +1,147 @@
...
+
+      # Constructor
+      #
+      # @param services [Array<SlpServiceClass::Service] list of services to show
+      def initialize(services = [])
+        super()
+
+        Yast.import "UI"
+        Yast.import "Label"
+        Yast.import "Report"
```
jreidinger: I expect import to be top level, it is one time operation, not needed for each initialization. It is like require

lslezak:
I suppose you will later move the new code somewhere to `yast2` or `yast2-slp` (this sounds better IMHO).

src/lib/registration/ui/service_selection_dialog.rb
lslezak: And it makes mocking in the tests easier because then you do not have to repeat the `import`s also in the test (you cannot mock undefined classes).

src/lib/registration/ui/service_selection_dialog.rb
```diff
@@ -0,0 +1,147 @@
...
+        @services = services
+      end
+
+      # Handler for the Ok button
+      #
+      # If no option was selected, a error message is shown.
+      def ok_handler
+        selected = Yast::UI.QueryWidget(Id(:services), :CurrentButton)
+        if !selected
+          Yast::Report.Error(_("No registration server selected."))
```
lslezak: If the code should be reusable we need to refactor these registration specific messages.

src/lib/registration/ui/service_selection_dialog.rb
```diff
@@ -0,0 +1,147 @@
...
+      # Dialog's initial content
+      #
+      # @return [Yast::Term] Content
+      def dialog_content
+        MarginBox(2, 0.5,
+          VBox(
+            # popup heading (in bold)
+            Heading(_("Local Registration Servers")),
+            VSpacing(0.5),
+            Label(_("Select a detected registration server from the list\n" \
```
lslezak: The same here, for a generic dialog the labels have to passed from outside.

src/lib/registration/ui/service_selection_dialog.rb
imobachgs: Yes, I'll need to pass a custom title/description in the installer's self-update client.

imobachgs:
@lslezak I'm not sure what's the best place for this dialog. It looks like 'registration' specific to me (after all, we're adding the SCC server to the list of services).

Perhaps we could have a `SelectionDialog` in yast2 (that one could be reusable) and implement this one (`ServiceSelectionDialog`) as a subclass. But I didn't go through that path because I was unsure if it would be useful.

Opinions?

lslezak:
It depends how much the new dialog will be registration independent. If it is completely registration independent and easily reusable then we could move it. If it still contains some registration specific parts or reusing it is difficult then I'd keep it here.

src/clients/discover_registration_services.rb
imobachgs: This client only returns the selected service (not stores it at all). So if the user goes back I expect a new instance of the client and the dialog to be used.

About handling only the abort case, this client returned `nil` if the user selects the "SCC" and I'm not changing that.

src/lib/registration/ui/service_selection_dialog.rb
imobachgs: Anyway, in this specific case, it looks like a quite general message (I don't see a reason to customize that one).

lslezak:
http://blog.ladslezak.cz/2016/06/06/overcommit/ and https://lists.opensuse.org/yast-devel/2016-06/msg00007.html :wink:

imobachgs:
Yes, you're right, but I haven't configured it in my laptop :)

lslezak:
It can be configured automatically: https://lists.opensuse.org/yast-devel/2016-06/msg00014.html

src/lib/registration/ui/service_selection_dialog.rb
lslezak: Then I'd probably make it even more generic (like `"No service selected."` or something like that).

src/lib/registration/ui/service_selection_dialog.rb
```diff
@@ -0,0 +1,154 @@
...
+              Id(index.to_s),
+              ::Registration::Helpers.service_description(service),
+              false
+            )
+          )
+        end
+
+        widgets.unshift(Left(RadioButton(Id("scc"),
+          # %s is the default SCC URL
+          _("SUSE Customer Center (%s)") % SUSE::Connect::YaST::DEFAULT_URL,
```
lslezak: This is a registration specific default. It depends if we want to make the dialog generic just for registration services or any other services as well.

imobachgs:
I guess I should do something with messages like [this one](https://github.com/yast/yast-registration/blob/cfc09cafb55fb228ef0298afb6b55149711ec739/src/lib/registration/connect_helpers.rb#L109). That method will take care of error handling but the message can be misleading when problem happens during installer update (and not during registration).

src/lib/registration/registration.rb
```diff
@@ -164,6 +165,25 @@ def migration_products(installed_products)
       migrations
     end
 
+    # Get the list of updates for a given product
+    #
+    # @param [Hash]           Product. If nil the base product will be used.
```
jreidinger: I think this really need description how hash should look like. Just by reading documentation I have no clue what to pass.

jreidinger: if it is documented elsewhere like in remote_product method, then at least please add `@see` to know where to look

src/lib/registration/ui/regservice_selection_dialog.rb
```diff
@@ -0,0 +1,91 @@
...
+      # * :cancel symbol if the SCC was canceled (pressing the 'cancel' button)
+      #
+      # @example Select the default SCC service
+      #   Registration::UI::SelectionServiceDialog.run(services) #=> :scc
+      #
+      # @example Select some SMT service
+      #   Registration::UI::SelectionServiceDialog.run(services)
+      #     #=> #<Yast::SlpServiceClass::Service...>
+      #
+      # @param services [Array<SlpServiceClass::Service] list of services to show
```
jreidinger: optional parameters not documented

src/lib/registration/ui/regservice_selection_dialog.rb
```diff
@@ -0,0 +1,91 @@
...
+      # @return [SlpServiceClass::Service,Symbol] selected service or symbol (:scc or :cancel)
+      #
+      # @see #run
+      def self.run(services:, heading: nil, description: nil)
+        new(services: services, heading: heading, description: description).run
+      end
+
+      # Constructor
+      #
+      # @param services [Array<SlpServiceClass::Service] list of services to show
```
jreidinger: same here

src/lib/registration/ui/regservice_selection_dialog.rb
```diff
@@ -0,0 +1,91 @@
...
+      # If no option was selected, a error message is shown.
+      #
+      # @see Yast::Dialogs::ServiceSelection#ok_handler
+      def ok_handler
+        selected = super
+        finish_dialog(:scc) if selected == scc_service
+      end
+
+      # Return the service description to be shown to the user
+      #
```
jreidinger: I expect here to documented parameter as it looks like super method do not handle string at all? so what can be passed?

jreidinger:
code wise LGTM, documentation wise there are some gaps

imobachgs:
@jreidinger Thank you! I'll improve the documentation and I'll search for a solution to [messages issue](https://github.com/yast/yast-registration/pull/263#issuecomment-235583016).

jreidinger:
@imobachgs maybe rephrase it like `connection to customer center failed`?

jreidinger:
LGTM

wfeldt:
lgtm

test/regservice_selection_dialog.rb
```diff
@@ -0,0 +1,103 @@
+require_relative "spec_helper"
```
mvidner: @imobachgs Only now I have noticed that this file does not end in _spec or _test (there are both kinds in this repo, ugh!) so it is not included in `rake test:unit`.

test/regservice_selection_dialog.rb
```diff
@@ -0,0 +1,103 @@
...
+    end
+
+    context "when the cancel button is pressed" do
+      it "returns :cancel" do
+        expect(Yast::UI).to receive(:UserInput).and_return(:cancel)
+        expect(dialog.run).to eq(:cancel)
+      end
+    end
+
+    it "sets default heading and description" do
```
mvidner:
```rb
textdomain "registration"
```
 should be here otherwise `y2tool y2makepot` complains

test/regservice_selection_dialog.rb
mvidner: Fixing that in #267

PR #264, opened 2016-07-24
[WIP] Free flag on smt (bsc#970990)

lslezak:
See bug https://bugzilla.suse.com/show_bug.cgi?id=970990

Unfortunately this won't work correctly in all cases. The SCC vs. SMT detection is currently done via the registration server URL.

This will work correctly with the SCC or an SMT server. However, this will break when using the SCC Proxy. In that case the URL is different than the SCC server but the specified server (the proxy) just forwards the registration to the real SCC. And SCC (in contrast to SMT) will require registration codes.

Unfortunately the SCC Proxy is used in openQA so we cannot ignore this use case.

The proposed solutions:

- Add a new flag in the server response. In addition to `free`, which defines whether the addon is paid, we could add a new flag (e.g. `regcode_required`) which would tell that the server requires a reg. code for this addon.  
*Unfortunately this would need a change on the SMT and SCC side. (Or at least on the SCC side, we could use default `false` for SMT when the value is missing in the response.)*
- Detect SMT via some API call which returns different value on SCC or SMT (or returns an different error). Or check the HTTP header?  
*The question is whether this is possible and how reliable it is...*
- Change the registration workflow - first try registering without any reg. code if it fails with "No reg. code" error ask user for the reg. code and try again.  
*Trying blindly registering an addon sounds a bit strange, we would need carefully check for the errors, esp. finding that it's just because of a missing reg. code would be tricky and likely fragile.*

Any other opinions, suggestions?

### DIFF
--- a/src/lib/registration/registration_ui.rb
+++ b/src/lib/registration/registration_ui.rb
@@ -205,9 +205,10 @@ module Registration
     #   user entered values
     # @return [Symbol]
     def register_addons(selected_addons, known_reg_codes)
-      # if registering only add-ons which do not need a reg. code (like SDK)
-      # then simply start the registration
-      if selected_addons.all?(&:free)
+      # if using an SMT server for registration (bsc#970990)
+      # or registering only add-ons which do not need a reg. code (like SDK)
+      # then simply start the registration without asking for the reg. codes
+      if !UrlHelpers.scc_url? || selected_addons.all?(&:free)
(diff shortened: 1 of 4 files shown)


lslezak:
I'm closing this as it currently seems we do not need this change.

I'll keep the branch so we can reopen if needed in the end.

See the discussion at https://bugzilla.suse.com/show_bug.cgi?id=970990.

PR #265, opened 2016-07-27
[L3 waiting for customer] Writable credentials dir

jreidinger:


lslezak:
The change itself is OK, but it is actually used to build DUD which fixes another libzypp DUD. It does not make much sense to release it for SP1 as it will not fix anything unless you apply the libzypp DUD.

If the PTF does not work or needs to be updated later we should probably better fix the original libzypp DUD instead of messing with the YaST workaround...

So I'm not merging the code, but I'll keep it for the future reference if needed.

PR #270, opened 2016-08-09
Removed workaround for making /etc/zypp directory writable (bsc#967828)

lslezak:
It has been fixed in the installation-images package - see https://github.com/openSUSE/installation-images/pull/126.

The `zypp_config_writable!` method is not needed anymore and can be removed.

- 3.1.183

### DIFF
--- a/package/yast2-registration.changes
+++ b/package/yast2-registration.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Tue Aug  9 12:59:25 UTC 2016 - lslezak@suse.cz
+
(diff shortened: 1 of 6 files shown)


imobachgs:
LGTM

kobliha:
It's nice to see some code removed, but having a test for this would be a benefit.

lslezak:
The `/etc/zypp` will be writable by default, see https://github.com/openSUSE/installation-images/pull/126. If that for some reason does not work then we will get failures (openQA or user bug reports) use very quickly as registration would not work at all.

PR #275, opened 2016-08-31
Style fix

jreidinger:
now it works in light and also dark theme

firstboot_registration_light
firstboot_registration_dark

### DIFF
--- a/package/yast2-registration.changes
+++ b/package/yast2-registration.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Wed Aug 31 09:10:41 UTC 2016 - jreidinger@suse.com
+
+- more robust check for installation qss for dark theme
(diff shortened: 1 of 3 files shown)


jreidinger:
note that screenshots use old style, ticks and arrow in firstboot is fixed with newer yast2-theme-SLE just tested it.

package/yast2-registration.changes
```diff
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Wed Aug 31 09:10:41 UTC 2016 - jreidinger@suse.com
+
+- more robust check for installation qss for dark theme
```
lslezak: `More robust check for installation dark theme` sounds better to me...

lslezak:
LGTM, thanks for the fix!

PR #277, opened 2016-09-19
Workaround for missing trailing characters (bsc#955156)

lslezak:
This is a workaround for an ncurses UI bug, the Qt UI displays the labels correctly.
See https://bugzilla.suse.com/show_bug.cgi?id=955156

### Notes

- This is for SP1 only as in SP2 we use a different UI layout and the bug does not happen there anymore.
- It's without a version update as this will be released later together with some other maintenance update (it is not urgent, it is just a cosmetic issue).

## The Original State

addon_selection_broken

## With the Workaround

addon_selection_fixed

### DIFF
--- a/package/yast2-registration.changes
+++ b/package/yast2-registration.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Mon Sep 19 12:08:54 UTC 2016 - lslezak@suse.cz
+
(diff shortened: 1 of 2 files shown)


shundhammer:
LGTM

mvidner:
LGTM

PR #278, opened 2016-10-03
Persistent beta state

jreidinger:


test/addon_selection_dialog_test.rb
```diff
@@ -17,6 +17,17 @@
     addon_reset_cache
   end
 
+  describe "#initialize" do
+    it "sets filter beta to previous state" do
```
lslezak: Just a better wording: "sets the beta filter to the previous state"

package/yast2-registration.changes
```diff
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Mon Oct  3 09:26:28 UTC 2016 - jreidinger@suse.com
+
+- remember value of filter beta checkbox (bnc#996891)
```
lslezak: And maybe we should rephrase it a bit so it's obvious that the stored setting is not permanent (it's lost after starting the module again in installed system).

Something like: `Remember the Beta filter value and set it when going back`

PR #279, opened 2016-10-03
Better handle invalid credentials (bsc#941427)

lslezak:
- Start the repository manager to remove the offending service/repository.
- Fixes [bsc#941427](https://bugzilla.suse.com/show_bug.cgi?id=941427)

### Notes

The debugging was a bit tricky, I was not able to reproduce the problem after manually invalidating (changing) the credentials stored in a *.service* file and clearing the libzypp cache with `zypper clear -a`.

It turned out that the service cache time stamp is stored directly in the *.service* file (`lrf_dat` key) and is not touched by that `zypper clear` call. By default the service cache is valid for 24 hours, so the invalid credentials might take effect after (up to) 24 hours!

### The Original Bug

When the credentials are not valid anymore and the service is refreshed when starting the registration module you will see this error:

yast2-failed_auth

Then whatever button you press you will see the generic exception handler:

yast2-original_failure

That's not user friendly.

### Initial Fix

At first I thought I could have added a new button (*Disable the Service*) into the authentication popup which would automatically disable the problematic service and user could continue in the workflow.

Unfortunately there are two issues:

- The [Authentication callback](https://github.com/yast/yast-yast2/blob/master/library/packages/src/modules/PackageCallbacks.rb#L2182) can be called not only from the service refresh, but also during package installation. In that context the *Disable the Service* button would not make sense. Unfortunately the context is not known for the callback code, we would need to track what is being called from the outside code. And that's fragile...
- The URL passed to the callback contains the real URL which failed (e.g. https://scc.suse.com/access/services/1106?cookies=0&credentials=SUSE_Linux_Enterprise_Server_12_x86_64), not the service URL (e.g. https://scc.suse.com/access/services/1106?credentials=SUSE_Linux_Enterprise_Server_12_x86_64). Matching the service URL and the failure URL would be tricky and again fragile (libzypp might add some more URL options or use different path in the future).

Therefore I implemented a simpler fix - when the repository initialization fails display the details about the failure and start the repository manager so the user could **manually** remove (disable) the offending service/repository.

### The Fix

After pressing any button in the authentication popup

yast2-failed_auth

you will see this new error popup with details:

yast2-failed_repos

After pressing *OK* the repository manager is started. Here you have to manually remove (or disable) the offending service/repository. After pressing *OK* there the registration module tries the repository initialization again. After pressing *Cancel* in the repository manager the registration module exits.

### DIFF
--- a/src/clients/scc.rb
+++ b/src/clients/scc.rb
@@ -58,6 +58,8 @@ module Yast
       end
     end
 
+  private
+
     # Print help in command line mode
     def print_help
       cmdline_description = {
--- a/src/clients/scc.rb
+++ b/src/clients/scc.rb
@@ -26,8 +26,13 @@
 require "yast"
 require "registration/sw_mgmt"
 
+# HTML escaping
+require "cgi/util"
+
 module Yast
   class SccClient < Client
+    include Yast::Logger
+
     Yast.import "CommandLine"
     Yast.import "Pkg"
     Yast.import "Report"
@@ -51,6 +56,16 @@ module Yast
           ::Registration::SwMgmt.init
 
           return WFM.call("inst_scc", WFM.Args)
+        rescue Registration::SourceRestoreError => e
+          # TRANSLATORS: Error message in RichText format, %s contains the details from libzypp
+          Report.LongError(_("<p>The repository initialization failed. " \
+            "Disable (or remove) the offending service or repository " \
+            "in the repository manager.</p><p>Details:</p><p>%s</p>") % CGI.escapeHTML(e.message))
+          ret = WFM.call("repositories", WFM.Args)
+          log.info "repository manager result: #{ret}"
+          # drop all loaded repos, force complete reloading
+          Pkg.SourceFinishAll
+          retry if ret == :next
         ensure
           Wizard.CloseDialog
         end
--- a/src/clients/scc.rb
+++ b/src/clients/scc.rb
@@ -42,13 +42,7 @@ module Yast
       textdomain "registration"
 
       if WFM.Args.include?("help")
-        cmdline_description = {
-          "id"   => "scc",
-          # Command line help text for the repository module, %1 is "SUSEconnect"
-          "help" => _("Use '%s' instead of this YaST module.") % "SUSEconnect"
-        }
-
-        CommandLine.Run(cmdline_description)
+        print_help
       else
         Wizard.CreateDialog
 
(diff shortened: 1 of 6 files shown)


src/clients/scc.rb
```diff
@@ -51,6 +56,16 @@ def main
...
+        rescue Registration::SourceRestoreError => e
+          # TRANSLATORS: Error message in RichText format, %s contains the details from libzypp
+          Report.LongError(_("<p>The repository initialization failed. " \
+            "Disable (or remove) the offending service or repository " \
+            "in the repository manager.</p><p>Details:</p><p>%s</p>") % CGI.escapeHTML(e.message))
+          ret = WFM.call("repositories", WFM.Args)
+          log.info "repository manager result: #{ret}"
+          # drop all loaded repos, force complete reloading
+          Pkg.SourceFinishAll
+          retry if ret == :next
```
jreidinger: well, at first, only SwMgmt.init should be enclosed and the second note, is that now method start to be quite long. So it should be shortened like:

```
def print_help
  cmdline_description = {
     "id"   => "scc",
     # Command line help text for the repository module, %1 is "SUSEconnect"
     "help" => _("Use '%s' instead of this YaST module.") % "SUSEconnect"
   }
 
   CommandLine.Run(cmdline_description)
end

def fix_repos
  # TRANSLATORS: Error message in RichText format, %s contains the details from libzypp
   Report.LongError(_("<p>The repository initialization failed. " \
     "Disable (or remove) the offending service or repository " \
     "in the repository manager.</p><p>Details:</p><p>%s</p>") % CGI.escapeHTML(e.message))
    ret = WFM.call("repositories", WFM.Args)
    log.info "repository manager result: #{ret}"
    # drop all loaded repos, force complete reloading
     Pkg.SourceFinishAll
     ret == :next
end

def main
       textdomain "registration"
 
       if WFM.Args.include?("help")
         print_help
       else
         Wizard.CreateDialog
 
         begin
           ::Registration::SwMgmt.init
 
           return WFM.call("inst_scc", WFM.Args)
         rescue Registration::SourceRestoreError => e
           retry if fix_repos
         ensure
           Wizard.CloseDialog
         end
end
```

btw that reminds me, that we should have safe block variant for Wizard.CreateDialog.

src/clients/scc.rb
```diff
@@ -37,25 +42,49 @@ def main
...
 
           return WFM.call("inst_scc", WFM.Args)
+        rescue Registration::SourceRestoreError => e
+          retry if fix_repositories(e.message)
         ensure
           Wizard.CloseDialog
         end
       end
     end
+
```
jreidinger: nice, almost perfect, just please add here `private` so it is clear what is public API and what is internal helper.

jreidinger:
LGTM

PR #281, opened 2016-10-05
Prepare maintenance update for SP2

lslezak:
This PR includes two backports to SP2:

- #279 
- #278 

No version change so far to avoid possible conflicts later, the version will be bumped just before submitting the maintenance update.


jreidinger:
LGTM

lslezak:
(Note: ignore the Jenkins failure, SP2 is not yet available in the public OBS.)

PR #283, opened 2016-11-09
SLE 12 SP2 merge

imobachgs:
Merge `merge_after_SP2_release` branch into `SLE-12-SP2`. It uses `3.1.191` because `3.1.189` and `3.1.190` already exist in `master` branch. We should bump `master` to `3.2.x`.

### DIFF
--- a/package/yast2-registration.changes
+++ b/package/yast2-registration.changes
@@ -8,7 +8,8 @@ Mon Oct  3 19:11:11 UTC 2016 - lslezak@suse.cz
 -------------------------------------------------------------------
 Mon Oct  3 09:26:28 UTC 2016 - jreidinger@suse.com
 
-- remember value of filter beta checkbox (bnc#996891)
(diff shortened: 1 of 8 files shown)


jreidinger:
well, version looks really strange and maybe invalid. I let @lslezak to decide if another two fixes in master should go there or not. We should not have in TW older version then in Leap/SP2

imobachgs:
You're right, it looks right. Do you want me to cherry-pick fixes for https://bugzilla.suse.com/show_bug.cgi?id=941427 and https://bugzilla.suse.com/show_bug.cgi?id=996891 ? If not, I'll just adjust the version number to `3.1.188.1`.

lslezak:
I'm fine with integrating the fixes.

Bumping the version in `master` is not critical (now), the registration module is *not* present in TW or Leap (it's SLE only).

Thanks Imo!

PR #284, opened 2016-11-10
Merge SLE 12 SP2 into master

imobachgs:


mvidner:
So the contents of the branches (SP2 and master) is supposed to be identical, right?

imobachgs:
Yes, the only difference is the submit target in the `Rakefile`. The point is to solve conflicts so when we merge changes in the future you won't need to take care about them.

imobachgs:
Thanks!

PR #288, opened 2016-11-24
Fix firstboot registration

jreidinger:


package/yast2-registration.changes
```diff
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Thu Nov 24 09:28:31 UTC 2016 - jreidinger@suse.com
+
+- allow user to go forward  in firstboot when system is already
```
lslezak: NP: double space in `forward  in`

PR #289, opened 2016-12-15
Use Docker at Travis, switch to the openSUSE based build

lslezak:
This is an experimental CI switch from the Ubuntu based build to the openSUSE Tumbleweed build using a Docker image. I'd like to see how it works in real life so lets switch some selected packages and see the result.

All features from the Ubuntu builds are kept, e.g. the code coverage reporting (coveralls) still works. See the [Travis Build log](https://travis-ci.org/yast/yast-registration/builds/184226734).

### Advantages

- Does not need packages built for Ubuntu (less maintenance for us, less troubles with building/publishing the packages by OBS)
- Closer matches our target environment (the same tooling)
- We can even build the rpm package from the sources (using plain `rpmbuild` to make it simple), that was not possible with Ubuntu, i.e. the `*.spec` file issues will be found now
- Can build Pull Requests (on Jenkins it's not trivial, @mvidner did not find any easy solution for that)
- You can easily reproduce the build locally, just run the same commands as in the `.travis.yml` file. That makes debugging much easier.

### Disadvantages

- A bit longer build, but not much (about +1 minute). It could be improved by building our custom Docker image with pre-installed YaST build dependencies (I'll look at this possibility later...)


.travis.sh
```diff
@@ -0,0 +1,20 @@
...
+/usr/bin/yardoc.ruby2.2
+# enable coveralls reports
+COVERAGE=1 CI=1 TRAVIS=1 rake test:unit
+
+# build the binary package locally, use plain "rpmbuild" to make it simple
+rake tarball
+cp package/* /usr/src/packages/SOURCES/
+rpmbuild -bb package/*.spec
+
+# test the %pre/%post script by installing/updating/removing the package
```
shundhammer: 👍 very good idea!

.travis.sh
```diff
@@ -0,0 +1,20 @@
...
+rpm -iv /usr/src/packages/RPMS/noarch/*.rpm
+rpm -Uv --force /usr/src/packages/RPMS/noarch/*.rpm
+rpm -ev yast2-registration
```
shundhammer: is that yast2-registration a leftovr from making this generic? Shouldn't we be able to get the package name from somewhere, say, the .spec file?

shundhammer:
LGTM, but I don't know anything about docker configuration, so I couldn't make a qualified statement about that.

.travis.sh
lslezak: Yes, getting the name somehow else would be better, I just did not want to spend much time here. It's more or less a prototype. I already several ideas for improvements...

PR #290, opened 2017-01-03
Travis/Docker update

lslezak:
# Docker/Travis Update

- Use a pre-built Docker image instead of building it locally
- Move the `.travis.sh` script directly to the Docker image to share it with other packages (moved [here](https://github.com/yast/docker-yast-ruby-tw/blob/master/yast-travis) and made more generic)

Additionally the obsolete `.pullreview.yml` file has been removed, the service is not used for some time...

# TODO

- I still expect some changes in the shared `yast-travis` script (it has not been tested with autotools based packages yet) and in the Docker image, let's use the agile approach and try it in the other packages and gradually fix the found issues...


jreidinger:
LGTM, just expect that that shell script need to be smart enough to find what different repos need.

PR #292, opened 2017-01-23
[RFC] Added Registration Code widget for CaaSP (Fate#322328)

teclator:
Just moved the current widget from the main PR: https://github.com/yast/yast-installation/pull/487 (that one uses a built-in fake registration widget)

Using this widget: https://github.com/yast/yast-installation/pull/504

If you want to know more about this:

- https://trello.com/c/XDgl2dj7
- https://fate.suse.com/322328

### DIFF
--- a/src/lib/registration/widgets/registration_code.rb
+++ b/src/lib/registration/widgets/registration_code.rb
@@ -70,7 +70,8 @@ module Registration
         end
 
         if !SwMgmt.find_base_product
-          return error(_("No base product found, skipping registration."))
+          Helpers.report_no_base_product
+          return false
         end
 
         log.info("Registering the system and the base product.")
--- a/src/lib/registration/widgets/registration_code.rb
+++ b/src/lib/registration/widgets/registration_code.rb
@@ -29,8 +29,6 @@ module Registration
     class RegistrationCode < CWM::InputField
       VALID_URL_SCHEMES = ["http", "https"].freeze
 
-      attr_accessor :registration_code, :registration, :reg_options
-
       def initialize
         textdomain "registration"
       end
@@ -41,22 +39,22 @@ module Registration
 
       # Initialize the widget with stored values
       def init
-        log.info("Already registered") if registered?
+        reg_code = options.reg_code.to_s
 
-        @options = Storage::InstallationOptions.instance
-        self.value = @options.custom_url
-        self.value = @options.reg_code if !@options.reg_code.to_s.empty?
+        self.value = reg_code.empty? ? options.custom_url || boot_url : options.reg_code
       end
 
       # Set registration options according to the value
       def store
         if valid_url?
-          @options.reg_code   = ""
-          @options.custom_url = value
+          options.reg_code   = ""
+          options.custom_url = value
         else
-          @options.reg_code   = value
-          @options.custom_url = default_url
+          options.reg_code   = value
+          options.custom_url = default_url
         end
+
+        register if !registered?
       end
 
       def registered?
@@ -64,7 +62,10 @@ module Registration
       end
 
       def register
-        return true if skip?
+        if skip?
+          log.info("Empty value, skipping registration")
+          return true
+        end
 
         log.info("Registering the system and the base product.")
         return false if !register_system_and_base_product
@@ -73,13 +74,38 @@ module Registration
         Storage::InstallationOptions.instance.base_registered = true
       end
 
+      def validate
+        return error(_("System already registered.")) if Registration.is_registered?
+
+        return true if skip?
+
+        return error(_("Not valid url.")) if url? & !valid_url?
+
+        true
+      end
+
     private
+
+      def error(message)
+        Yast::Popup.Error(message)
+
+        false
+      end
+
+      def url?
+        uri = URI(value)
+        uri.scheme ? true : false
+      rescue URI::InvalidURIError
+        false
+      end
 
       def valid_url?
         return false if value.to_s.empty?
 
         uri = URI(value)
         VALID_URL_SCHEMES.include?(uri.scheme)
+      rescue URI::InvalidURIError
+        false
       end
 
       # run the system and the base product registration
@@ -120,6 +146,10 @@ module Registration
       # Skip registration if the value is empty or nil
       def skip?
         value.to_s.empty?
+      end
+
+      def options
+        Storage::InstallationOptions.instance
       end
     end
   end
--- a/src/lib/registration/widgets/registration_code.rb
+++ b/src/lib/registration/widgets/registration_code.rb
@@ -1,0 +1,35 @@
+# encoding: utf-8
+
+# ------------------------------------------------------------------------------
+# Copyright (c) 2017 SUSE LLC
+#
+#
+# This program is free software; you can redistribute it and/or modify it under
+# the terms of version 2 of the GNU General Public License as published by the
+# Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+# FOR A PARTICULAR PURPOSE. See the GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along with
+# this program; if not, contact SUSE.
+#
+# To contact SUSE about this file by physical or electronic mail, you may find
+# current contact information at www.suse.com.
+# ------------------------------------------------------------------------------
+
+require "yast"
+require "cwm/widget"
+
+module Widgets
+  class RegistrationCode < CWM::InputField
+    def initialize
+      textdomain "registration"
+    end
+
+    def label
+      _("Registration Code or SMT Server URL")
+    end
+  end
+end
--- a/src/lib/registration/widgets/registration_code.rb
+++ b/src/lib/registration/widgets/registration_code.rb
@@ -74,10 +74,13 @@ module Registration
           return false
         end
 
+
         log.info("Registering the system and the base product.")
+        # Error reports and logs about the registration are mostly handled
+        # by ConnectHelpers.catch_registration_errors and used by instances
+        # of RegistrationUI.
         return false if !register_system_and_base_product
 
-        log.info("System registered with success.")
         Storage::InstallationOptions.instance.base_registered = true
       end
 
--- a/src/lib/registration/widgets/registration_code.rb
+++ b/src/lib/registration/widgets/registration_code.rb
@@ -70,8 +70,7 @@ module Registration
         end
 
         if !SwMgmt.find_base_product
-          log.info("Not base product found, skipping registration.")
-          return true
+          return error(_("No base product found, skipping registration."))
         end
 
         log.info("Registering the system and the base product.")
--- a/src/lib/registration/widgets/registration_code.rb
+++ b/src/lib/registration/widgets/registration_code.rb
@@ -21,15 +21,106 @@
 
 require "yast"
 require "cwm/widget"
+require "uri"
+require "registration/registration_ui"
 
-module Widgets
-  class RegistrationCode < CWM::InputField
-    def initialize
-      textdomain "registration"
-    end
+module Registration
+  module Widgets
+    class RegistrationCode < CWM::InputField
+      VALID_URL_SCHEMES = ["http", "https"].freeze
 
-    def label
-      _("Registration Code or SMT Server URL")
+      attr_accessor :registration_code, :registration, :reg_options
+
+      def initialize
+        textdomain "registration"
+      end
+
+      def label
+        _("Registration Code or SMT Server URL")
+      end
+
+      # Initialize the widget with stored values
+      def init
+        log.info("Already registered") if registered?
+
+        @options = Storage::InstallationOptions.instance
+        self.value = @options.custom_url
+        self.value = @options.reg_code if !@options.reg_code.to_s.empty?
+      end
+
+      # Set registration options according to the value
+      def store
+        if valid_url?
+          @options.reg_code   = ""
+          @options.custom_url = value
+        else
+          @options.reg_code   = value
+          @options.custom_url = default_url
+        end
+      end
+
+      def registered?
+        Registration.is_registered?
+      end
+
+      def register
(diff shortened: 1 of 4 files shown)


mvidner:
Thanks!

If we merged this as is, it would confuse people because we say we added a widget but we don't mention that the widget does not do anything.

We should either say "(a dummy so far)" in .changes, or implement some real registration, even if not perfect.

src/lib/registration/widgets/registration_code.rb
```diff
@@ -0,0 +1,126 @@
...
+
+      def label
+        _("Registration Code or SMT Server URL")
+      end
+
+      # Initialize the widget with stored values
+      def init
+        log.info("Already registered") if registered?
+
+        @options = Storage::InstallationOptions.instance
```
jreidinger: why Storage? Please use full name, it is very confusing

teclator:
We could reuse most of the logic implemented in this class, by now I have just copied some of the methods and modified a bit

https://github.com/yast/yast-registration/blob/master/src/lib/registration/ui/base_system_registration_dialog.rb

src/lib/registration/widgets/registration_code.rb
```diff
@@ -0,0 +1,126 @@
...
+          @options.reg_code   = value
+          @options.custom_url = default_url
+        end
+      end
+
+      def registered?
+        Registration.is_registered?
+      end
+
+      def register
```
jreidinger: we discuss it now with @jsrain that this should be in fact `store` method, so it will be registered here. And if registration failed, then just show popup "Registration failed, please register later." and go forward

src/lib/registration/widgets/registration_code.rb
```diff
@@ -0,0 +1,126 @@
...
+      end
+
+    private
+
+      def valid_url?
+        return false if value.to_s.empty?
+
+        uri = URI(value)
+        VALID_URL_SCHEMES.include?(uri.scheme)
+      end
```
jreidinger: in fact this should be handled in validation. We need to recognize what looks like registration code, what is empty and also recognize valid url for smt and in other case report problem.
and this should be only `url?` to recognize if value is url or not.

src/lib/registration/widgets/registration_code.rb
```diff
@@ -0,0 +1,126 @@
...
+        textdomain "registration"
+      end
+
+      def label
+        _("Registration Code or SMT Server URL")
+      end
+
+      # Initialize the widget with stored values
+      def init
+        log.info("Already registered") if registered?
```
jreidinger: in fact this should never happen. as we do not allow going back, if I get workflow now.

src/lib/registration/widgets/registration_code.rb
```diff
@@ -0,0 +1,126 @@
...
+require "cwm/widget"
+require "uri"
+require "registration/registration_ui"
+
+module Registration
+  module Widgets
+    class RegistrationCode < CWM::InputField
+      VALID_URL_SCHEMES = ["http", "https"].freeze
+
+      attr_accessor :registration_code, :registration, :reg_options
```
jreidinger: not used except reg_options and I think that it also should not be visible.

src/lib/registration/widgets/registration_code.rb
```diff
@@ -0,0 +1,126 @@
...
+        _("Registration Code or SMT Server URL")
+      end
+
+      # Initialize the widget with stored values
+      def init
+        log.info("Already registered") if registered?
+
+        @options = Storage::InstallationOptions.instance
+        self.value = @options.custom_url
+        self.value = @options.reg_code if !@options.reg_code.to_s.empty?
```
jreidinger: this is quite confusing. I think more readable is 
```
self.value = @options.reg_code.to_s.empty? ?  @options.custom_url : @options.reg_code
```

src/lib/registration/widgets/registration_code.rb
```diff
@@ -0,0 +1,126 @@
...
+        log.info("System registered with success.")
+        Storage::InstallationOptions.instance.base_registered = true
+      end
+
+    private
+
+      def valid_url?
+        return false if value.to_s.empty?
+
+        uri = URI(value)
```
lslezak: Beware, invalid URL throws the `URI::InvalidURIError` exception here... :boom:

src/lib/registration/widgets/registration_code.rb
```diff
@@ -0,0 +1,126 @@
...
+      end
+
+      # Default registration server
+      #
+      # The boot_url takes precedence over the SUSE::Connect default
+      # one.
+      #
+      # @return [String] URL for the registration server
+      def default_url
+        boot_url || SUSE::Connect::Config.new.url
```
jreidinger: ok, so what happen if boot_url is used and then this field kept empty? I expect that nothing is written, which is a bit inconsistent.

jreidinger: maybe use this boot_url as default value?

src/lib/registration/widgets/registration_code.rb
teclator: yep, I does not copy all the logic yet, I just submit it to do start discussing ;) as commented many logic could be usable so if we prefer to duplicate by now I will finish adding the reusable logic here and then think in move to a specialized class

src/lib/registration/widgets/registration_code.rb
teclator: we should initialize the field with the boot_url or the custom_url in case of empty reg_code, I have to check it better

src/lib/registration/widgets/registration_code.rb
teclator: sure

src/lib/registration/widgets/registration_code.rb
teclator: yep, I started used them, but then just used the `Registration::Storage::Storage::InstallationOptions.instance` as https://github.com/yast/yast-registration/blob/master/src/lib/registration/ui/base_system_registration_dialog.rb was using, so this could be removed.

src/lib/registration/widgets/registration_code.rb
teclator: yep, just in case of update and that is out of the scope by now

src/lib/registration/widgets/registration_code.rb
teclator: yep, by now I just took care of custom urls, if empty then skip, and if not treat it as a registration code, which means registration will fail and report it to the user, but is better to handle it by the validation method and only allow `VALID_URL_SCHEMES`

src/lib/registration/widgets/registration_code.rb
teclator: then, perfect, I was worried if it could be launched by other method trying to register so decided to move the responsibility out of the store method.

src/lib/registration/widgets/registration_code.rb
```diff
@@ -73,13 +74,38 @@ def register
         Storage::InstallationOptions.instance.base_registered = true
       end
 
+      def validate
+        return error(_("System already registered.")) if Registration.is_registered?
```
jreidinger: so when system is already registered you basically are stuck here, because you cannot do anything about it.

src/lib/registration/widgets/registration_code.rb
```diff
@@ -73,13 +74,38 @@ def register
...
+
+        return true if skip?
+
+        return error(_("Not valid url.")) if url? & !valid_url?
```
jreidinger: I suggest to write some help here, like "Only supported schemas are http and https"?

src/lib/registration/widgets/registration_code.rb
teclator: hehe, true 👍

src/lib/registration/widgets/registration_code.rb
```diff
@@ -41,30 +39,33 @@ def label
 
       # Initialize the widget with stored values
       def init
-        log.info("Already registered") if registered?
+        reg_code = options.reg_code.to_s
 
-        @options = Storage::InstallationOptions.instance
-        self.value = @options.custom_url
-        self.value = @options.reg_code if !@options.reg_code.to_s.empty?
+        self.value = reg_code.empty? ? options.custom_url || boot_url : options.reg_code
```
lslezak: NP: I'd prefer having parenthesis here just for the readability reasons, `? ?` and `|| :` groups are bit harder to read.

```ruby
(reg_code.empty?) ? (options.custom_url || boot_url) : options.reg_code
```

But if Rubocop is against then keep this :wink:

src/lib/registration/widgets/registration_code.rb
lslezak: In case of already registered system we should simply skip registration and display just a warning.

But as user normally cannot go back after registration (AFAIK the installation is started even after failed registration) this check is actually irrelevant. But let's keep it just to be on the safe side...

src/lib/registration/widgets/registration_code.rb
lslezak: Maybe simpler:

```ruby
(url? & !valid_url?) ? error(_("Not valid url.")) : true
```

src/lib/registration/widgets/registration_code.rb
teclator: Yep, I will just log in case of already registered, but also in that case we could show the base product label instead of the widget as already registered in the content of the casp overview client.

src/lib/registration/widgets/registration_code.rb
teclator: rubocop is not fine with the first parenthesis, but second one could help readability so changing it

src/lib/registration/widgets/registration_code.rb
```diff
@@ -67,6 +64,16 @@ def register
           return true
         end
 
+        if Registration.is_registered?
+          log.info("The system is already registered so skipped registration.")
+          return true
+        end
+
+        if !SwMgmt.find_base_product
```
teclator: This check is just a safe check, we could rely that always that the widget is used there exists a base_product but just be safer.

src/lib/registration/widgets/registration_code.rb
```diff
@@ -67,6 +64,16 @@ def register
           return true
         end
 
+        if Registration.is_registered?
+          log.info("The system is already registered so skipped registration.")
```
teclator: We could warn the user, just logged by now because it should not happen and also we could just check it out of the widget rendering the label of the product instead of a input field.

src/lib/registration/widgets/registration_code.rb
```diff
@@ -0,0 +1,166 @@
...
+          return true
+        end
+
+        if Registration.is_registered?
+          log.info("The system is already registered so skipped registration.")
+          return true
+        end
+
+        if !SwMgmt.find_base_product
+          log.info("Not base product found, skipping registration.")
```
lslezak: I think you should be more aggressive here, at least display an error. Or even throw an exception and abort the installation.

Missing base product issue can happen only in the development phase, usually as a result of some bug in the product definition or as an OBS bug. I have already seen such bug in the past, this can really happen.

And if that problem happens we should not be silent, the release managers should be aware of this issue ASAP. Hiding this bug may cause some problems later.

teclator: yep, but was not sure if handle here or in the inst_casp_overview  client so was the reason to my comment and also to add here to at least not go over it.

I saw your checks at the ins_scc client:

https://github.com/yast/yast-registration/blob/master/src/clients/inst_scc.rb#L182

src/lib/registration/widgets/registration_code.rb
```diff
@@ -70,8 +70,7 @@ def register
         end
 
         if !SwMgmt.find_base_product
-          log.info("Not base product found, skipping registration.")
-          return true
+          return error(_("No base product found, skipping registration."))
```
lslezak: You can reuse the [Registration::Helpers.report_no_base_product](https://github.com/yast/yast-registration/blob/327ab34c020a89f8b7e3f4bff55deea82e457237/src/lib/registration/helpers.rb#L232) method here to avoid new translations.

teclator: Reuse then !!  ;) thnx

src/lib/registration/widgets/registration_code.rb
```diff
@@ -0,0 +1,166 @@
...
+      def store
+        if valid_url?
+          options.reg_code   = ""
+          options.custom_url = value
+        else
+          options.reg_code   = value
+          options.custom_url = default_url
+        end
+
+        register
```
jreidinger: if register report false, we ignore it. Does it show popup itself, or it should be showed here?

teclator: Report errors are mainly handled here:

https://github.com/yast/yast-registration/blob/master/src/lib/registration/connect_helpers.rb#L60

which is called by RegistrationUI https://github.com/yast/yast-registration/blob/master/src/lib/registration/registration_ui.rb#L59

jreidinger: si please write it to code, as it is not much obvious, that error reporting already happen and it should be also tested what happen when wrong key is used and with other problems ( like network down ).

src/lib/registration/widgets/registration_code.rb
```diff
@@ -61,20 +61,19 @@ def store
       def register
         if skip?
           log.info("Empty value, skipping registration")
-          return true
+          return false
```
jreidinger: why such change? we do not care about return value, so at least document what you want to archieve with it

teclator: Basically be consistent with the API only returns true if the registration success, I modified it based on test that I'm written, I thought that is more consistent and expected.

teclator: As you commented as we are not taking take care of the output does not affect, but in case of already registered then register will not be called or just return false.

jreidinger: so please document it in method documentation, as it is not obvious. One approach is true only if it is really registered, second approach is return false only if something goes wrong.

PR #297, opened 2017-01-30
Created desktop file for running "add extensions" directly

mchf:
This implementation is done using desktop file only, however it doesn't work in ncurses

### DIFF
--- a/package/yast2-registration.changes
+++ b/package/yast2-registration.changes
@@ -1,4 +1,12 @@
 -------------------------------------------------------------------
+Mon Jan 30 20:05:56 UTC 2017 - mfilka@suse.com
+
+- fate#321044
+  - provide a shortcut for running "Add System Extensions/Modules"
+    dialog directly from control centre.
+- 3.2.2
+
+-------------------------------------------------------------------
 Fri Jan 27 17:18:29 UTC 2017 - kanderssen@suse.com
 
 - Added Registration::Widgets::RegistrationCode (fate#322328).
--- a/src/desktop/add_extensions.desktop
+++ b/src/desktop/add_extensions.desktop
@@ -3,6 +3,7 @@ Type=Application
 Categories=Settings;System;Qt;X-SuSE-YaST;X-SuSE-YaST-Software;
 
 X-SuSE-YaST-Call=scc
+X-SuSE-YaST-Argument=select_extensions
 X-SuSE-YaST-RootOnly=yes
 X-SuSE-YaST-Group=Software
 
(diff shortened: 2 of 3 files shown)


src/desktop/add_extensions.desktop
```diff
@@ -0,0 +1,15 @@
...
+X-SuSE-YaST-Call=scc
+X-SuSE-YaST-Argument=select_extensions
+X-SuSE-YaST-RootOnly=yes
+X-SuSE-YaST-Group=Software
+
+Icon=yast-product-registration
+Exec=/sbin/yast2 scc select_extensions
+
+Name=Add System Extensions/Modules
+GenericName=Add System Extensions/Modules
```
lslezak: I'm not sure if using the slash is OK in the label. IIRC the old styleguide was against this and suggested using `and` or `or` words instead.

I have checked the other items and `and` is used several times, no item with slash. So I'd change it to be more consistent. (Fix all places, it's repeated several times in the file.)

src/desktop/add_extensions.desktop
```diff
@@ -0,0 +1,15 @@
+[Desktop Entry]
+Type=Application
+Categories=Settings;System;Qt;X-SuSE-YaST;X-SuSE-YaST-Software;
+
+X-SuSE-YaST-Call=scc
+X-SuSE-YaST-Argument=select_extensions
+X-SuSE-YaST-RootOnly=yes
+X-SuSE-YaST-Group=Software
+
+Icon=yast-product-registration
```
lslezak: I'd use another icon here, this is primarily for adding new extensions, the registration dialog is displayed only when the system is not registered yet.

I'd prefer the same icon as for the add-on module, i.e. `Icon=yast-addon`.

src/desktop/add_extensions.desktop
mchf: just note: it worked fine in gnome even ncurses. I'll change it.

src/desktop/add_extensions.desktop
mchf: no problem, I thought that someone from UX will check that before release.

package/yast2-registration.changes
```diff
@@ -1,4 +1,12 @@
 -------------------------------------------------------------------
+Mon Jan 30 20:05:56 UTC 2017 - mfilka@suse.com
+
+- fate#321044
+  - provide a shortcut for running "Add System Extensions/Modules"
```
lslezak: `or`...

mchf: here it refers to the dialog title ....

mchf:
thanks for review

PR #298, opened 2017-01-30
[WIP] Addon desktop workaround

mchf:
This is workaround which works even in ncurses, but I don't like it. So, I'd go for this one only if https://github.com/yast/yast-registration/pull/297 cannot be fixed other way (TBD)


mchf:
ok, as usual workaround is not needed shortly after exposing that ;-)

PR #299, opened 2017-02-14
Get update list errors

imobachgs:
* `Registration#get_updates_list` won't catch errors anymore. I think this method should be responsible for just asking the API.
* Other option could be just adding a new argument (`handle_errors` or something like that) which determines whether error handling should be performed or not.

The error is handled in the [UpdateRepositoryFinder class](https://github.com/yast/yast-installation/pull/522/commits/65860510f385bacb027fd6e928262be8881d3f81#diff-bf4d34fcdb7355d3d66880e4bf99d1c4R284).

### DIFF
--- a/package/yast2-registration.changes
+++ b/package/yast2-registration.changes
@@ -1,4 +1,12 @@
 -------------------------------------------------------------------
+Tue Feb 14 10:00:41 UTC 2017 - igonzalezsosa@suse.com
+
+- Registration#get_updates_list won't catch errors anymore because
+  that's something tightly coupled to the user interface. They
+  should be handled outside that method (bsc#1025251)
+- 3.2.3
+
+-------------------------------------------------------------------
 Mon Jan 30 20:05:56 UTC 2017 - mfilka@suse.com
 
 - fate#321044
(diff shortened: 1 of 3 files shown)


teclator:
I like decoupling it from the interface, but in that case the exceptions should be handled in some place, and currently are not.

https://github.com/yast/yast-installation/pull/522/files#diff-bf4d34fcdb7355d3d66880e4bf99d1c4R42 or I'm missing something?

imobachgs:
Yes, they're handled in this PR: https://github.com/imobachgs/yast-installation/pull/1 :) I know it's confusing, so I've merged that PR into the original one and you can see now all changes: https://github.com/yast/yast-installation/pull/522.

teclator:
Thnx for clarify it ;)

imobachgs:
Thanks!

mvidner:
Please mention the other PR  (or even a particular commit) in a commit message, so that it is easier to find.

imobachgs:
@mvidner The error is now handled here: https://github.com/yast/yast-installation/pull/522/commits/65860510f385bacb027fd6e928262be8881d3f81#diff-bf4d34fcdb7355d3d66880e4bf99d1c4R284

PR #300, opened 2017-02-24
Display more addons in text mode (bsc#1024464)

lslezak:
See bug https://bugzilla.novell.com/show_bug.cgi?id=1024464

The dialog was originally designed for much less addons, unfortunately the number of addon grew over time and it exceeded the original limit for the text mode UI (18 addons max, currently there are 19 addons).

#### Notes

- The graphical (Qt) version is not affected as the maximum number of displayed addons is much bigger (28 addons).
- That means the graphical layout was kept untouched to avoid possible regressions.
- Only SP1 is affected, in SP2 we use a scrollable `RichText` widget

### The Original Dialog

addons_broken

The dialog looks OK, but actually the *Web and Scripting Module* is missing :hushed: 

## Fix

addons_80x25

- Decreased the size of the *Details* description
- Increased the number of displayed addons from 18 to 22
- The *Web and Scripting Module* is displayed properly
- There is even space for 3 more addons (so hopefully we do not run into the same problem soon...)

### Bigger Screen

The number of addons in one column is now computed dynamically depending on the current screen size.

#### Bigger Screen Size

If the there is enough free space then the dialog might be even switched to one column and the labels are not truncated anymore:

addons_97x35

#### Even Bigger Screen Size

For even bigger screens the *Details* description size is increased:

addons_105x41

### DIFF
--- a/package/yast2-registration.changes
+++ b/package/yast2-registration.changes
@@ -1,7 +1,7 @@
 -------------------------------------------------------------------
 Thu Mar  2 12:54:20 UTC 2017 - lslezak@suse.cz
 
-- Fixed layout of the "Extension and Module Selection" dialog
+- Fixed the layout of the "Extension and Module Selection" dialog
   in text mode to make all items visible (bsc#1024464)
 - 3.1.166.4
 
--- a/package/yast2-registration.changes
+++ b/package/yast2-registration.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Thu Mar  2 12:54:20 UTC 2017 - lslezak@suse.cz
+
+- Fixed layout of the "Extension and Module Selection" dialog
+  in text mode to make all items visible (bsc#1024464)
+- 3.1.166.4
+
+-------------------------------------------------------------------
 Mon Oct 17 11:21:34 UTC 2016 - lslezak@suse.cz
 
 - Recommend yast2-migration package to install it by default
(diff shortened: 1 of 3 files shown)


package/yast2-registration.changes
```diff
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Thu Mar  2 12:54:20 UTC 2017 - lslezak@suse.cz
+
+- Fixed layout of the "Extension and Module Selection" dialog
+  in text mode to make all items visible (bsc#1024464)
```
teclator: I'm not the best to comment about wording ... probably Martin could comment but couldn't be better `making all the items visible`

teclator:
Maybe some constant could be used for the MIN and MAX values, but in general looks good specially taking in account that is only for SP1.

And also thanks not only for the fix but also for the nice documented PR with screenshots and comparisons about what was happening and how looks now I think that it makes it more easy to get an idea about the current problem and be in context more quickly as well it is more fun :P 🥇

lslezak:
Yes, I was thinking about some constants, but as this is just for SP1 I preferred a simple and small fix.

package/yast2-registration.changes
mvidner: This is fine, but in"Fixed layout of the ... dialog", "fixed" is an adjective (Pevné rozvržení dialogu) whereas we want "Fixed **the** layout of the ... dialog".

PR #301, opened 2017-03-05
Be explicit about 'Details' box not being localized (bsc#1025846)

okurz:
By adding a simple notice it should be clarified that the product description
is english and is expected to be so.

### DIFF
--- a/package/yast2-registration.changes
+++ b/package/yast2-registration.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Mon Mar 20 12:28:45 UTC 2017 - okurz@suse.com
+
+- Be explicit about 'Details' box not being localized (bsc#1025846)
+- 3.2.6
+
+-------------------------------------------------------------------
 Fri Mar 17 09:23:03 UTC 2017 - gsouzadossantos@suse.com
 
 - Remember the state of the checkbox when leaving the dialog
--- a/src/lib/registration/ui/addon_selection_base_dialog.rb
+++ b/src/lib/registration/ui/addon_selection_base_dialog.rb
@@ -102,7 +102,7 @@ module Registration
           Left(CheckBox(Id(:filter_beta), Opt(:notify),
             _("&Hide Beta Versions"), check_filter)),
           addons_box,
-          Left(Label(_("Details"))),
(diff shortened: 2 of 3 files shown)


kalabiyau:
I think until proper description translation we could help customer - being more explicit about the fact. So, 👍  from me.

lslezak:
Just a note, bug https://bugzilla.suse.com/show_bug.cgi?id=714541 (mentioned in the commit message) is about an unrelated problem. Please fix that.

lslezak:
@kwwii not sure whether `Not Translated` would be better than `English Only`. Your opinion?

okurz:
LOL, I don't have *any* idea how I came up with this bugref. Corrected in commit message and PR subject

lslezak:
After thinking about it a bit I came to the conclusion that the "English Only" variant is better than "Not Translated" as this might be confusing if you install in English. In that case the word "translation" does not make sense...

src/lib/registration/ui/addon_selection_base_dialog.rb
```diff
@@ -96,7 +96,7 @@ def content
           Left(CheckBox(Id(:filter_beta), Opt(:notify),
             _("&Filter Out Beta Versions"), FILTER_BETAS_INITIALLY)),
           addons_box,
-          Left(Label(_("Details"))),
+          Left(Label(_("Details (english only)"))),
```
lslezak: Please use capital `E`, the language names always use a capital letter in English.

And please add a `.changes` entry with the `bsc` reference and bump the version to `3.2.6` in the spec file (manually or run `rake version:bump`).

Thank you!

package/yast2-registration.changes
```diff
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Mon Mar 20 12:28:45 UTC 2017 - okurz@suse.com
+
+- Be explicit about 'Details' box not being localized (bsc#1025846)
+- 3.2.4
```
lslezak: Typo, it's `3.2.6` now...

okurz:
done

lslezak:
Merged, thanks for your contribution! :+1:

PR #303, opened 2017-03-13
Remember the state of the checkbox when leaving the dialog (bsc#1026155)

gilsonsouza:


src/lib/registration/ui/addon_selection_base_dialog.rb
```diff
@@ -71,7 +71,13 @@ def addon_widget_id(addon)
       # @param [Boolean] enable true for filtering beta releases
       def filter_beta_releases(enable)
         self.class.filter_beta = enable
-        @addons = enable ? @all_addons.reject(&:beta_release?) : @all_addons
+        if enable
+          unregistered_addons = @all_addons.reject(&:registered?)
```
jreidinger: well, not so nice nice, as it is not unregistered, but not yet registered, so maybe available_addons?

src/lib/registration/ui/addon_selection_base_dialog.rb
```diff
@@ -71,7 +71,13 @@ def addon_widget_id(addon)
...
+          @addons = @all_addons.select(&:registered?)
```
jreidinger: btw as you split by registered, you can have just one call - http://apidock.com/ruby/Enumerable/partition
so something like
```
@addons, available_addons = @all_addons.partition(&:registered?)
```

jreidinger:
logic wise it looks ok, also where is some test case?

src/lib/registration/ui/addon_selection_base_dialog.rb
gilsonsouza: I agree that unregistered is a little ambiguous since it can give the idea of deregistered. But available_addons can give the idea that this variable has all available add-ons, instead of only the ones not yet registered and deregistered (if some day we implement deregister).

package/yast2-registration.changes
```diff
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Mon Mar 13 11:12:19 UTC 2017 - gsouzadossantos@suse.com
+
+- Remember the state of the checkbox when leaving the dialog 
+  (bsc#1026155) and correct the filter to always show installed
+  addons, even if they are beta versions.
+
```
imobachgs: Just a minor one: please, add a line with the new version number like we do https://github.com/gilsonsouza/yast-registration/blob/169fdad67e062ebf7b5b3a99f174f20ce781f4e5/package/yast2-registration.changes#L13.

src/lib/registration/ui/addon_selection_base_dialog.rb
gilsonsouza: This approach breaks the **initialize** test because it returns a NilClass object into the available_addons

jreidinger:
On Tue, 14 Mar 2017 04:00:38 -0700
gilsonsouza <notifications@github.com> wrote:

> gilsonsouza commented on this pull request.
> 
> 
> 
> > @@ -71,7 +71,13 @@ def addon_widget_id(addon)  
>        # @param [Boolean] enable true for filtering beta releases
>        def filter_beta_releases(enable)
>          self.class.filter_beta = enable
> -        @addons = enable ? @all_addons.reject(&:beta_release?) :
> @all_addons
> +        if enable
> +          unregistered_addons = @all_addons.reject(&:registered?)
> +          @addons = @all_addons.select(&:registered?)
> 
> This approach breaks the **initialize** test because it returns a
> NilClass object into the available_addons
> 

? it should not return nil unless there is a bug.

```
irb(main):003:0> nils, valid = [0,2].partition(&:nil?)
=> [[], [0, 2]]
irb(main):004:0> nils
=> []
irb(main):005:0> valid
=> [0, 2]
```

PR #304, opened 2017-03-15
Remember the state of the checkbox when leaving the dialog (bsc#1026155)

gilsonsouza:


test/addon_selection_dialog_test.rb
```diff
@@ -62,6 +70,32 @@
       expect(wrapped_addon.selected?).to eq true
     end
 
+    context "when beta versions are not filtered" do
+      subject(:dialog) { described_class.new(registration) }
+      let(:filter_beta) { false }
+
+      it "sets the filter as 'enabled' in the UI" do
```
imobachgs: I know it was my mistake, but we could change it to "sets the filter as not checked in the UI" so it looks more clear. What do you think? Same for the following context.

src/lib/registration/ui/addon_selection_base_dialog.rb
```diff
@@ -71,7 +71,12 @@ def addon_widget_id(addon)
       # @param [Boolean] enable true for filtering beta releases
       def filter_beta_releases(enable)
         self.class.filter_beta = enable
-        @addons = enable ? @all_addons.reject(&:beta_release?) : @all_addons
+        if enable
+          @addons, available_addons = @all_addons.partition(&:registered?)
```
imobachgs: I would extend the test to check that the content of the list. You could just add some add-on and just check if it was added or not to the list.

You could do that checking if the call to [Yast::UI::ChangeWidget](https://github.com/gilsonsouza/yast-registration/blob/af91a26a5290066d2d395d3b0591dd2b4e12bc2c/src/lib/registration/ui/addon_selection_base_dialog.rb#L251) call includes that add-on in the last parameter. What's your opinion?

So you could have: 

```ruby
context "when beta versions are not filtered" do
  it "sets the filter as not checked in the UI" do
     # the test you already has
  end

  it "displays beta add-ons" do
    # DISCLAIMER: not tested :)
    # the new test could be something like this:
    expect(Yast::UI).to receive(:ChangeWidget)
      .with(Yast::Term.new(:id, :items), :Value, /name of the beta add-on/) # you can use a regular expression to match with a string
    expect(Yast::UI).to receive(:UserInput).and_return(:next)
    dialog.run
  end
```

Sorry for noticing before about it.

test/addon_selection_dialog_test.rb
```diff
@@ -62,6 +70,72 @@
...
+      it "display registered beta add-ons" do
+        allow(addon).to receive(:beta_release?).and_return(true)
+        allow(addon).to receive(:registered?).and_return(true)
+        allow(Registration::Addon).to receive(:find_all).and_return([addon])
+        expect(subject).to receive(:RichText).with(Yast::Term.new(:id, :items), /#{addon.name}/)
+        allow(subject).to receive(:RichText).and_call_original
+        expect(Yast::UI).to receive(:UserInput).and_return(:next)
+        dialog.run
+      end
+
```
imobachgs: Please, just remove this blank line. I'm wonder why rubocop didn't complain.

PR #307, opened 2017-03-30
Added migration warnings when not installed but registered (Fate#320534).

teclator:
Trello Card: https://trello.com/c/CyjL1Was

Workflow added:

1. **Start YaST Migration**
2. **Check registered products not installed (continue in case of empty)**
    - The user is able to try to install the **release package** if available.
    - The user is able to synchronize the products deactivating the ones not installed from SCC.
    - Just continue to the migration selection summary.
3. **Find installed products and migrations for them**
    - if not solved inconsistencies we ask the user about the not installed addons allowing him to add them to the current migration (one per one).
    - Show the summary of the available migrations.
    - In case of a third party add-on (not migrations available) we recommend to use manual repository edition.
4. **Continue to the migration proposal**
    - In case of **abort**, the user is asked again about adding all the registered but not installed products to the downgrade and synchronization (not removing the products if accepted)
    - Only the already activated products are downgraded and synchronized (fixing a bug with third party addons).

### New warning dialog (not installed but registered products)

Basically it let you install the release package of the removed product if still present (it is what zypper migration do) or synchronize your current installed products (deactivating the ones not installed but registered).

**animated gif**
screencast_install_packages

**screenshot**
screenshot from 2017-03-30 21-25-13

**Continue without sync or install**
continue

**Migration selection summary**
screenshot from 2017-03-31 09-44-12

### Abort migration (only already activated products)

screenshot_sles12_2017-03-23_10 55 01

Previously it always failed in case of third party addons because SCC hasn't got a vision about them. If we only downgrade and synchronize the already activated addons then there is no problem. We also ask about registered but not installed products to be added to the downgrade and synchronization not deactivating them in case of confirm.

**screenshot**
screenshot from 2017-03-30 23-51-38

### DIFF
--- a/src/lib/registration/registration_ui.rb
+++ b/src/lib/registration/registration_ui.rb
@@ -287,6 +287,27 @@ module Registration
       options.install_updates
     end
 
+    # Ask the user if wants to also rollback the registered but not
+    # installed addons, in case of accept, it returns the addons list.
+    #
+    # @return [Array<OpenStruct>] registered but not installed addons if
+    #   accept or an empty array if not.
+    def registered_addons_to_rollback
+      get_available_addons
+
+      addon_names = Addon.registered_not_installed.map(&:friendly_name)
+
+      return [] if addon_names.empty?
+
+      msg = _("The addons listed below are registered but not installed: \n\n%s\n\n" \
+              "Would you like to rollback also them? If not, they will be deactivated. ")
+      if Yast::Popup.YesNo(msg % addon_names.join("\n"))
+        Addon.registered_not_installed
+      else
+        []
+      end
+    end
+
     private
 
     attr_accessor :registration
--- a/src/lib/registration/registration_ui.rb
+++ b/src/lib/registration/registration_ui.rb
@@ -305,7 +305,7 @@ module Registration
 
       return [] if addons.empty?
 
-      addon_names = addons.map {|a| a["display_name"] }
+      addon_names = addons.map { |a| a["display_name"] }
 
       msg = _("The addons listed below are registered but not installed: \n\n%s\n\n" \
               "Would you like to downgrade also them in the registration server? \n" \
--- a/src/lib/registration/registration_ui.rb
+++ b/src/lib/registration/registration_ui.rb
@@ -295,14 +295,23 @@ module Registration
     def registered_addons_to_rollback
       get_available_addons
 
-      addon_names = Addon.registered_not_installed.map(&:friendly_name)
+      addons =
+        Addon.registered_not_installed.map do |addon|
+          ret = addon.to_h
+          ret["display_name"]    = addon.friendly_name
+          ret["version_version"] = addon.version
+          ret
+        end
 
-      return [] if addon_names.empty?
+      return [] if addons.empty?
+
+      addon_names = addons.map {|a| a["display_name"] }
 
       msg = _("The addons listed below are registered but not installed: \n\n%s\n\n" \
-              "Would you like to rollback also them? If not, they will be deactivated. ")
+              "Would you like to downgrade also them in the registration server? \n" \
+              "If not they will be deactivated. ")
       if Yast::Popup.YesNo(msg % addon_names.join("\n"))
-        Addon.registered_not_installed
+        addons
       else
         []
       end
--- a/src/lib/registration/registration_ui.rb
+++ b/src/lib/registration/registration_ui.rb
@@ -307,14 +307,15 @@ module Registration
 
       addon_names = addons.map { |a| a["display_name"] }
 
+      # TRANSLATORS: Popup question, add registered but not installed addons to
+      # the list of products that will be downgraded.
+      # %s are all the product names splited by '\n' e.g
+      # "SUSE Linux Enterprise Server 12\nSUSE Enterprise Storage 1 x86_64"
       msg = _("The addons listed below are registered but not installed: \n\n%s\n\n" \
               "Would you like to downgrade also them in the registration server? \n" \
-              "If not they will be deactivated. ")
-      if Yast::Popup.YesNo(msg % addon_names.join("\n"))
-        addons
-      else
-        []
-      end
+              "If not they will be deactivated. ") % addon_names.join("\n")
+
+      Yast::Popup.YesNo(msg) ? addons : []
     end
 
     private
--- a/src/lib/registration/ui/migration_repos_workflow.rb
+++ b/src/lib/registration/ui/migration_repos_workflow.rb
@@ -224,14 +224,14 @@ module Registration
         # TRANSLATORS: Popup question, merge this addon that are registered but not
         # installed to the current migration products list.
         # %s is an addon friendly name, e.g 'SUSE Enterprise Storage 2 x86_64'
-        msg = "The '%s' extension is registered but not installed.\n" \
+        msg = _("The '%s' extension is registered but not installed.\n" \
               "If you accept it will be added for be installed, in other case " \
               "it will be unregistered at the end of the migration.\n\n" \
-              "Do you want to add it?"
+              "Do you want to add it?")
 
         addons =
           Addon.registered_not_installed.each_with_object([]) do |addon, result|
-            if Yast::Popup.YesNoHeadline(addon.friendly_name, _(msg % addon.friendly_name))
+            if Yast::Popup.YesNoHeadline(addon.friendly_name, (msg % addon.friendly_name))
               result << SwMgmt.remote_product(addon.to_h)
             end
           end
--- a/src/lib/registration/ui/migration_repos_workflow.rb
+++ b/src/lib/registration/ui/migration_repos_workflow.rb
@@ -221,6 +221,9 @@ module Registration
         # load the extensions to merge the registered but not installed extensions
         Addon.find_all(registration)
 
+        # TRANSLATORS: Popup question, merge this addon that are registered but not
+        # installed to the current migration products list.
+        # %s is an addon friendly name, e.g 'SUSE Enterprise Storage 2 x86_64'
         msg = "The '%s' extension is registered but not installed.\n" \
               "If you accept it will be added for be installed, in other case " \
               "it will be unregistered at the end of the migration.\n\n" \
--- a/src/lib/registration/ui/migration_repos_workflow.rb
+++ b/src/lib/registration/ui/migration_repos_workflow.rb
@@ -23,6 +23,7 @@ require "registration/url_helpers"
 require "registration/ui/wizard_client"
 require "registration/ui/migration_selection_dialog"
 require "registration/ui/migration_repos_selection_dialog"
+require "registration/ui/not_installed_products_dialog"
 
 module Registration
   module UI
@@ -55,6 +56,8 @@ module Registration
       #
       # - if the system is not registered ask the user to register it first
       #   (otherwise abort the online migration)
+      # - check registered but not installed products allowing the user to
+      #   syncronize them (skipped in case of not found)
       # - find all installed products
       # - query registered addons from the server
       # - ask the registration server for the available product migrations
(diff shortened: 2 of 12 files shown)


src/lib/registration/ui/not_installed_products_dialog.rb
```diff
@@ -0,0 +1,161 @@
...
+          MinWidth(80,
+            MinHeight(15,
+              VWeight(25, RichText(Id(:summary), not_installed_addons_summary))
+            )
+          ),
+          ButtonBox(
+            PushButton(Id(:cancel), Opt(:key_F9, :cancelButton), Yast::Label.AbortButton),
+            # FIXME: Maybe we could remove this option and just warn the user
+            PushButton(Id(:install), _("Ins&tall addons")),
+            PushButton(Id(:sync), _("&Synchronize")),
```
lslezak: The help label says *Synchronize your products if you want to deactivate them at your registration server* so I'd use *Deactivate* button label here and describe what does it mean in the help label.

src/lib/registration/ui/not_installed_products_dialog.rb
```diff
@@ -0,0 +1,161 @@
...
+          Yast::UI.SetFocus(:next)
+          ui = Yast::UI.UserInput
+          log.info "User input: #{ui}"
+          case ui
+          when :install
+            not_installed = []
+            Addon.registered_not_installed.map do |addon|
+              begin
+                Yast::Popup.Feedback(RegistrationUI::CONTACTING_MESSAGE,
+                  _("Installing %s release package") % addon.identifier) do
```
lslezak: NP: Add `TRANSLATORS:` comment describing what `%s` means, possibly with an example.

src/lib/registration/ui/not_installed_products_dialog.rb
```diff
@@ -0,0 +1,161 @@
...
+
+                  Yast::Pkg.ResolvableInstall(addon.identifier, :product)
+                  Yast::Pkg.PkgSolve(true)
+                  Yast::Pkg.PkgCommit(0)
+                end
+              rescue
+                not_installed << addon.identifier
+              end
+            end
+            Yast::Popup.Error(_("These addons were not installed:\n\n%s") %
```
lslezak: Again, a short description for `%s` would help the translators.

src/lib/registration/ui/migration_selection_dialog.rb
```diff
@@ -179,24 +179,44 @@ def update_details
...
 
       # create a product summary for the details widget
       # @return [String] product summary
       def product_summary(product, installed_product)
-        product_name = CGI.escapeHTML(product.friendly_name)
         log.info "creating summary for #{product} and #{installed_product}"
 
+        if !product
+          product_name = CGI.escapeHTML(SwMgmt.product_label(installed_product))
+          return _("The registration server does not offer migrations for Product " \
```
lslezak: A short description for `%s` would help the translators.

src/lib/registration/ui/migration_repos_workflow.rb
```diff
@@ -193,18 +208,30 @@ def find_products
...
+      #
+      # @return [Array<Hash>] installed products and addons selected to be installed
       def merge_registered_addons
         # load the extensions to merge the registered but not installed extensions
         Addon.find_all(registration)
 
-        addons = Addon.registered_not_installed.map(&:to_h).map do |addon|
-          SwMgmt.remote_product(addon)
-        end
+        msg = "The '%s' extension is registered but not installed.\n" \
```
lslezak: A short description for `%s` would help the translators.

src/lib/registration/registration_ui.rb
```diff
@@ -287,6 +287,36 @@ def install_updates?
...
+        end
+
+      return [] if addons.empty?
+
+      addon_names = addons.map { |a| a["display_name"] }
+
+      msg = _("The addons listed below are registered but not installed: \n\n%s\n\n" \
+              "Would you like to downgrade also them in the registration server? \n" \
+              "If not they will be deactivated. ")
+      if Yast::Popup.YesNo(msg % addon_names.join("\n"))
```
lslezak: NP: may be the triple operator would be more readable:

```ruby
Yast::Popup.YesNo(msg % addon_names.join("\n")) ? addons : []
```

Or maybe 

```ruby
msg = _("...") % addon_names.join("\n")
Yast::Popup.YesNo(msg) ? addons : []
```

src/lib/registration/ui/migration_repos_workflow.rb
```diff
@@ -193,18 +208,33 @@ def find_products
...
         # load the extensions to merge the registered but not installed extensions
         Addon.find_all(registration)
 
-        addons = Addon.registered_not_installed.map(&:to_h).map do |addon|
-          SwMgmt.remote_product(addon)
-        end
+        # TRANSLATORS: Popup question, merge this addon that are registered but not
+        # installed to the current migration products list.
+        # %s is an addon friendly name, e.g 'SUSE Enterprise Storage 2 x86_64'
+        msg = "The '%s' extension is registered but not installed.\n" \
```
lslezak: The `_()` translation mark is missing here, gettext will not include the text the generated POT file.

To use the translation multiple times use `N_()` function here...

src/lib/registration/ui/migration_repos_workflow.rb
```diff
@@ -193,18 +208,33 @@ def find_products
...
+              "If you accept it will be added for be installed, in other case " \
+              "it will be unregistered at the end of the migration.\n\n" \
+              "Do you want to add it?"
+
+        addons =
+          Addon.registered_not_installed.each_with_object([]) do |addon, result|
+            if Yast::Popup.YesNoHeadline(addon.friendly_name, _(msg % addon.friendly_name))
```
lslezak: ... and `_(msg) % addon.friendly_name` here.

*Never* modify the string *before* translating it otherwise gettext will not find the translation for it.

See more details here: https://github.com/yast/yast.github.io/blob/master/doc/localization.md#translating-constants

src/lib/registration/ui/migration_selection_dialog.rb
```diff
@@ -179,24 +179,46 @@ def update_details
...
       def product_summary(product, installed_product)
-        product_name = CGI.escapeHTML(product.friendly_name)
         log.info "creating summary for #{product} and #{installed_product}"
 
+        if !product
+          product_name = CGI.escapeHTML(SwMgmt.product_label(installed_product))
+          # TRANSLATORS: Summary message in rich text format
+          # %s is a product name, e.g. "SUSE Linux Enterprise Server 12 SP1 x86_64"
+          return _("The registration server does not offer migrations for Product " \
+                   "<b>%s</b> so it will <b>stay unchanged</b> . We recommend you " \
```
lslezak: NP: Space before `.`.

src/lib/registration/ui/not_installed_products_dialog.rb
```diff
@@ -0,0 +1,196 @@
...
+        end
+      end
+
+      # Install the not installed products
+      # @return [Array] list of not installed (failed) products, empty list on success
+      def install_products
+        not_installed = []
+
+        Addon.registered_not_installed.each do |addon|
+          Yast::Popup.Feedback(RegistrationUI::CONTACTING_MESSAGE,
```
lslezak: I do not think we need yet another popup, normally there should be a package installation progress displayed. If not then we could just run the `inst_rpmcopy.rb` client instead of the `Yast::Pkg.PkgCommit(0)` call.

src/lib/registration/ui/migration_repos_workflow.rb
lslezak: Actually you do not have to use `N_()` in this case:

```
msg = _("...")
...
... msg % addon.friendly_name ...
```

teclator: yep 👍 I already did this change :) just not taking in account transformation when changed the code. so it will be `msg = _("...")` and that is all

lslezak:
Almost LGTM, just add a changes entry and bump the version.

src/lib/registration/ui/not_installed_products_dialog.rb
```diff
@@ -126,10 +126,7 @@ def install_products
 
             result = Yast::Pkg.PkgCommit(0)
             # success?
-            if result && result[1].empty?
-              Yast::PackagesUI.show_update_messages(result)
```
teclator: It is not available in SLE-12-GA

src/lib/registration/ui/not_installed_products_dialog.rb
```diff
@@ -0,0 +1,225 @@
...
+      # Given an URL return the concatenation of host and path removing all the
+      # end slashes.
+      #
+      # FIXME: This method could be moved to UrlHelpers or to URL module.
+      #
+      # @param [String] url to parse
+      # @return [String] url without params
+      def no_params_url(url)
+        uri = URI.parse(url)
+        "#{uri.host}#{uri.path.gsub(/\/+$/, "")}"
```
lslezak: I'd rather compare also the other URL properties, it's unlikely that the same URL just with a different scheme (or user/password) will be in the system, but it's better to be on the safe side:

```ruby
uri.query = nil
uri.path.gsub!(/\/+\z/, "")
```

Then you can compare the URI objects directly.

src/lib/registration/ui/not_installed_products_dialog.rb
lslezak: Please add it as commented out with a reason, it will help when doing other backports later.

lslezak:
And also add commented out `PackagesUI.show_update_messages` call, just to know why it is missing.

src/lib/registration/ui/not_installed_products_dialog.rb
```diff
@@ -0,0 +1,234 @@
...
+# You should have received a copy of the GNU General Public License along with
+# this program; if not, contact SUSE.
+#
+# To contact SUSE about this file by physical or electronic mail, you may find
+# current contact information at www.suse.com.
+# ------------------------------------------------------------------------------
+
+require "yast"
+require "registration/sw_mgmt"
+require "registration/helpers"
```
lslezak: Add also `require "uri"` as it is also used here. (It is loaded transitively so it works, but that's fragile and will break when it's removed.)

teclator: Totally agree, just missed it, added 👍

src/lib/registration/ui/not_installed_products_dialog.rb
```diff
@@ -0,0 +1,235 @@
...
+
+      def initialize
+        textdomain "registration"
+
+        self.registration = Registration.new(UrlHelpers.registration_url)
+        self.registration_ui = RegistrationUI.new(registration)
+      end
+
+      def run
+        Yast::Popup.Feedback(RegistrationUI::CONTACTING_MESSAGE, _("Checking product conflicts")) do
```
lslezak: NP: Um, it's a bit misleading IMHO, it's not about conflicts between the products, "Checking registration status" or something like that would be better I think.

src/lib/registration/ui/not_installed_products_dialog.rb
```diff
@@ -0,0 +1,235 @@
...
+        VBox(
+          MinWidth(80,
+            MinHeight(15,
+              VWeight(25, RichText(Id(:summary), not_installed_addons_summary))
+            )
+          ),
+          ButtonBox(
+            PushButton(Id(:cancel), Opt(:key_F9, :cancelButton), Yast::Label.AbortButton),
+            # FIXME: Maybe we could remove this option and just warn the user
+            PushButton(Id(:install), _("Ins&tall addons")),
```
lslezak: NP: "addons" might also a bit confusing, officialy we provide "modules" or "extensions". Simple "Install Products" is generic and covers both possibilities.

src/lib/registration/ui/not_installed_products_dialog.rb
```diff
@@ -0,0 +1,235 @@
...
+      def not_installed_addons_summary
+        # TRANSLATORS: A RichText warning about all the products registered but
+        #   not installed. (1/2)
+        summary = _("<p>The addons listed below are registered but not installed: </p>")
+
+        summary << "<ul>#{not_installed_addon_names.map { |a| "<li>#{a}</li>" }.join("")}</ul>"
+
+        # TRANSLATORS: A RichText warning about all the products registered but
+        #   not installed. (2/2)
+        summary << _("<p><b>Synchronize</b> your products if you want to <b>deactivate</b> " \
```
lslezak: The `Synchronize` button has been renamed, update the help text accordingly.

src/lib/registration/ui/not_installed_products_dialog.rb
```diff
@@ -0,0 +1,235 @@
...
+                     "them at your registration server.</p>")
+
+        summary
+      end
+
+      # It updates the summary of registered but not installed products
+      # updating also Addon cache.
+      def update_summary
+        log.info "Updating summary"
+        Yast::Popup.Feedback(RegistrationUI::CONTACTING_MESSAGE, _("Updating summary")) do
```
lslezak: You can reuse the proposed "Checking registration status" label here. Save the translators work, and also we do not have to reveal all internals to the user.

src/lib/registration/ui/not_installed_products_dialog.rb
```diff
@@ -56,7 +58,7 @@ def initialize
       end
 
       def run
-        Yast::Popup.Feedback(RegistrationUI::CONTACTING_MESSAGE, _("Checking product conflicts")) do
+        Yast::Popup.Feedback(RegistrationUI::CONTACTING_MESSAGE, REGISTRATION_CHECK_MSG) do
```
lslezak: You need to use `_()` with `N_()` as it just marks the text as translatable but does not actually do the translation. See http://yastgithubio.readthedocs.io/en/latest/localization/#translating-constants for more details.

src/lib/registration/ui/not_installed_products_dialog.rb
```diff
@@ -168,7 +170,7 @@ def not_installed_addons_summary
       # updating also Addon cache.
       def update_summary
         log.info "Updating summary"
-        Yast::Popup.Feedback(RegistrationUI::CONTACTING_MESSAGE, _("Updating summary")) do
+        Yast::Popup.Feedback(RegistrationUI::CONTACTING_MESSAGE, REGISTRATION_CHECK_MSG) do
```
lslezak: The same here.

PR #312, opened 2017-04-11
Merge branch 'SLE-12-SP2' into SLE-12-SP2-CASP

teclator:
This merges #307 into CaaSP branch

### DIFF
--- a/package/yast2-registration.changes
+++ b/package/yast2-registration.changes
@@ -1,10 +1,24 @@
 -------------------------------------------------------------------
+Mon Apr 10 07:38:45 WEST 2017 - knut.anderssen@suse.com
+
+- Online Migration (fate#320534)
+  - Added step for check registered but not installed addons to the
+    migration workflow. In case of existence, allow the user to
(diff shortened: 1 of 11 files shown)


lslezak:
Do we actually need this in CaaSP? I mean the `yast2-migration` package is not included in the CaaSP (actually there is no YaST package installed in the target system). And CaaSP does not support any upgrade, it will always be a new installation.

teclator:
@lslezak I was not sure at all, as you commented although it is SLE-12-SP2 based, it will not support upgrades and then the yast2-migration package should be added. 

But in that case then all SLE-12-SP2 changes will not be in SLE-12-CASP breaking the current rule.. well there are cases and cases.

lslezak:
The rule is ["when submit to CASP, merge to SP2"](https://trello.com/c/Oarngtp2) - we do not need it in CASP and in SP2 it is already there. In the end the CASP branch will be merged to SP2.

So I think we can close it without merging.

teclator:
@lslezak ok let's do so by now

PR #315, opened 2017-04-28
Copy certificate to /etc/pki/trust/anchors instead of

schubi2:
+  /usr/share/pki/trust/anchors

scc_auto:Write will be called quite after the scc_auto:Import function in inst_autosetup. 
So there is no installed system available at that point.


schubi2:
wrong tree

PR #316, opened 2017-04-28
Copy certificate to /etc/pki/trust/anchors instead of

schubi2:
+  /usr/share/pki/trust/anchors
scc_auto:Write will be called quite after the scc_auto:Import function in inst_autosetup.
So there is no installed system available at that point.

### DIFF
--- a/package/yast2-registration.changes
+++ b/package/yast2-registration.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Fri Apr 28 11:11:22 CEST 2017 - schubi@suse.de
+
+- AY inst-sys: Copy certificate to /etc/pki/trust/anchors instead of
+  /usr/share/pki/trust/anchors wich is read only (bnc#103466).
+- 3.1.193
+
+-------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


package/yast2-registration.changes
```diff
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Fri Apr 28 11:11:22 CEST 2017 - schubi@suse.de
+
+- AY inst-sys: Copy certificate to /etc/pki/trust/anchors instead of
+  /usr/share/pki/trust/anchors wich is read only (bnc#103466).
```
imobachgs: typo: `which`

imobachgs: Is the bug number ok?

src/clients/scc_auto.rb
```diff
@@ -252,7 +252,7 @@ def import_certificate(url)
       end
 
       Popup.Feedback(_("Importing SSL Certificate"), cert.subject_name) do
-        cert.import_to_system
+        cert.import
```
imobachgs: 👍

PR #319, opened 2017-05-08
Don't crash in case of an invalid linuxrc registration url is provided (bsc#1035908)

teclator:
This PR is related to yast/yast-installation#563

### Screenshot during manual installation

screenshot_sled12sp3_2017-05-08_12 03 41

### AutoYast fail currently catch it as a connection error.

screenshot_sle12sp3_2017-05-08_14 50 16

### DIFF
--- a/package/yast2-registration.changes
+++ b/package/yast2-registration.changes
@@ -1,4 +1,12 @@
 -------------------------------------------------------------------
+
+Mon May  8 13:27:09 UTC 2017 - knut.anderssen@suse.com
+
+- Don't crash if the regurl provided by linuxrc is invalid using
+  the one provided by the control file as fallback (bsc#1035908).
+- 3.2.10
+
+-------------------------------------------------------------------
 Fri Apr 28 11:11:22 CEST 2017 - schubi@suse.de
 
 - AY inst-sys: Copy certificate to /etc/pki/trust/anchors instead of
@@ -30,7 +38,6 @@ Mon Apr 10 07:38:45 WEST 2017 - knut.anderssen@suse.com
 - 3.2.7
 
 -------------------------------------------------------------------
-
 Mon Mar 20 12:28:45 UTC 2017 - okurz@suse.com
 
 - Be explicit about 'Details' box not being localized (bsc#1025846)
--- a/package/yast2-registration.changes
+++ b/package/yast2-registration.changes
@@ -2,7 +2,7 @@
 
 Mon May  8 13:27:09 UTC 2017 - knut.anderssen@suse.com
 
-- Don't crash if the regurl provided by linuxrc is invalid using
+- Don't crash if the regurl provided by linuxrc is invalid, use
   the one provided by the control file as fallback (bsc#1035908).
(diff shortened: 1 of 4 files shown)


src/lib/registration/ui/base_system_registration_dialog.rb
```diff
@@ -229,7 +229,19 @@ def default_url
       #
       # @return [String,nil] URL for the registration server; nil if not given.
       def boot_url
-        @boot_url ||= UrlHelpers.boot_reg_url
+        return nil unless UrlHelpers.boot_reg_url
+
+        unless valid_custom_url?(UrlHelpers.boot_reg_url)
+          Yast::Report.Error(
+            # TRANSLATORS: Wrong url for registration provided, %s is an URL.
+            _("The registration URL provided by Linuxrc is not valid.\n\n" \
```
lslezak: End users might not be familiar with `Linuxrc`, for them it's an URL written at the boot command line.

src/lib/registration/ui/base_system_registration_dialog.rb
```diff
@@ -229,7 +229,19 @@ def default_url
...
+              "URL: %s\n\nThe default one will be used.") % UrlHelpers.boot_reg_url
```
lslezak: Do we know the default URL at this point? If so it would be a good idea to print it here.

package/yast2-registration.changes
```diff
@@ -1,4 +1,12 @@
 -------------------------------------------------------------------
+
+Mon May  8 13:27:09 UTC 2017 - knut.anderssen@suse.com
+
+- Don't crash if the regurl provided by linuxrc is invalid using
```
lslezak: Better wording: ... if ... is invalid, use the one...

Again, avoid linuxrc term.

src/lib/registration/ui/base_system_registration_dialog.rb
teclator: Well, it was really influenced by default_url but it really don't know if we are using the default one or not, so not make sense in that method, which points to the question if we should report it there or in the default one.

src/lib/registration/ui/base_system_registration_dialog.rb
teclator: Modified

package/yast2-registration.changes
teclator: Updated

src/lib/registration/ui/base_system_registration_dialog.rb
teclator: The check was moved to the default method, so we could show it at this point although it is shown in the next screen where shows the registration URL

teclator:
@lslezak, shall I create an alternative PR to also ask for a new URL for this scenario?

lslezak:
> @lslezak, shall I create an alternative PR to also ask for a new URL for this scenario?

No, the URL should be already fixed by the self-update popup displayed earlier, in AY mode the error popup is enough.

PR #331, opened 2017-08-30
Wrap connect helper errors text (bsc#1055643)

teclator:
- It depends on yast/yast-yast2#618

## Before the fix
registration - message too long to read

## After the fix
registration - wrap message

### DIFF
--- a/src/lib/registration/connect_helpers.rb
+++ b/src/lib/registration/connect_helpers.rb
@@ -175,7 +175,7 @@ module Registration
       return error if !details || details.empty?
 
       # %s are error details
-      error + "\n\n" + (_("Details: %s") % wrap_text(details))
+      error + "\n\n" + wrap_text(_("Details: %s") % details)
     end
 
     def self.ssl_error_details(cert)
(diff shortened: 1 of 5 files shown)


src/lib/registration/connect_helpers.rb
```diff
@@ -173,7 +175,7 @@ def self.error_with_details(error, details)
       return error if !details || details.empty?
 
       # %s are error details
-      error + "\n\n" + (_("Details: %s") % details)
+      error + "\n\n" + (_("Details: %s") % wrap_text(details))
```
teclator: Details: should be used as prepend text

test/connect_helpers_spec.rb
```diff
@@ -42,8 +47,11 @@ def api_error(code: 400, headers: {}, body: "")
     it "reports an error with details on timeout" do
       details = _("Make sure that the registration server is reachable and\n" \
         "the connection is reliable.")
+
+      expect(subject).to receive(:wrap_text).with("Details: #{details}", 76)
```
lslezak: The value `76` depends on the `80` value above, it would be nice to have `let(:screen_width) { 80 }` and use it mocking and the test.

teclator: :1:

PR #332, opened 2017-09-12
Fixed internal error when cloning configuration

mchf:
bnc#1056318


mchf:
thanks for the review

PR #334, opened 2017-10-03
Initial support for the SLE12 -> SLE15 upgrade + cleanup + addon dumper

lslezak:
## Summary

- Initial support for the SLE12 -> SLE15 upgrade, honor the target mount point prefix so it works in inst-sys ([fate#323163](https://fate.suse.com/323163))
- Do not crash when baseproduct symlink does not exist in installed system
- Dump the available addons to a separate file (`/var/log/YaST2/registration_addons.yml`) instead of the `y2log` to make it better readable, added simple dump reader for easier debugging on a different system

## Screencast

 scc_upgrade

## Notes

- This is a proof of concept, it uses the SP level migration API on the SCC side, that's why in the demo we can only upgrade from SLE12-SP2 to SLE12-SP3 using the SLE15 installation medium
- The full functionality will need changes on the SCC side (new API)


lslezak:
Additionally moved the `inst_scc` client code to `lib/`, this fixes counting the code coverage. (The code executed through `WFM.call` is not properly marked as executed.)

src/lib/registration/addon.rb
```diff
@@ -45,6 +45,19 @@ def find_all(registration)
...
+        if File.writable?("/var/log/YaST2")
+          require "yaml"
+          header = "# see " \
+            "https://github.com/yast/yast-registration/tree/master/devel/dump_reader.rb\n" \
+            "# for an example how to read this dump file\n"
+          File.write("/var/log/YaST2/registration_addons.yml",
+            header + @cached_addons.to_yaml)
+        end
+
+        @cached_addons
```
jreidinger: this method start to be too big.

src/lib/registration/clients/inst_scc.rb
```diff
@@ -0,0 +1,347 @@
...
+      textdomain "registration"
+      import_modules
+
+      first_run
+
+      @selected_addons = ::Registration::Storage::InstallationOptions.instance.selected_addons
+
+      initialize_regcodes
+
+      media_workflow? ? ::Registration::UI::MediaAddonWorkflow.run(WFM.Args[1]) : start_workflow
```
jreidinger: this smells for me. Why not having it as two classes one for offline and one for online? with common ancestor.

jreidinger: in general changing behavior when different caller call you is bad.

src/lib/registration/clients/inst_scc.rb
```diff
@@ -0,0 +1,347 @@
...
+    def first_run
+      return unless ::Registration::Storage::Cache.instance.first_run
+
+      ::Registration::Storage::Cache.instance.first_run = false
+
+      return unless Stage.initial && ::Registration::Registration.is_registered?
+
+      ::Registration::Helpers.reset_registration_status
+    end
+  end unless defined?(InstSccClient)
```
jreidinger: this is not needed for code in "lib" only for code in "clients"

src/lib/registration/clients/inst_scc.rb
lslezak: OK, but that would require non trivial change and also changes outside the package. So I just added a `FIXME` to fix that later.

PR #335, opened 2017-10-05
Improve addon ui

jreidinger:
demonstration of functionality

- [x] unit tests

better icon is in https://github.com/yast/yast-theme/pull/83

test

### DIFF
--- a/src/lib/registration/addon.rb
+++ b/src/lib/registration/addon.rb
@@ -94,15 +94,21 @@ module Registration
       # @return [Array<Addon>] list of addons, where the first one is
       #   the one based on root and rest is its children
       def create_addon_with_deps(root)
-        root_addon = Addon.new(root)
-        result = [root_addon]
+        # to_process is array of pairs, where first is pure addon to process and second is
+        # its dependency. Currently SUSE::Connect structure have only one dependency.
+        to_process = [[root, nil]]
+        processed = Set.new
+        result = []
 
-        (root.extensions || []).each do |ext|
-          # FIXME: endless loop if circular dependency between addons happen
-          child = create_addon_with_deps(ext)
-          result.concat(child)
-          child.first.depends_on = root_addon
-          root_addon.children << child.first
+        to_process.each do |(ext, dependency)|
+          next if processed.include?(ext)
+          processed << ext
+          addon = Addon.new(ext)
+          result << addon
+          addon.depends_on = dependency
+          (ext.extensions || []).each do |ext|
+            to_process << [ext, addon]
+          end
         end
 
         result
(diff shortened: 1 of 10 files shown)


src/lib/registration/addon.rb
```diff
@@ -84,25 +83,70 @@ def registered_not_installed
...
-          child.first.depends_on = root_addon
-          root_addon.children << child.first
+        # to_process is array of pairs, where first is pure addon to process and second is
+        # its dependency. Currently SUSE::Connect structure have only one dependency.
+        to_process = [[root, nil]]
+        processed = Set.new
+        result = []
+
+        to_process.each do |(pure, dependency)|
+          next if processed.include?(pure)
```
lslezak: NP: I'd add a comment saying "avoid infinite loop when there is a cyclic dependency", just you can quickly see that this case is also handled here.

src/lib/registration/addon.rb
```diff
@@ -84,25 +83,70 @@ def registered_not_installed
...
+          processed << pure
+          addon = Addon.new(pure)
+          result << addon
+          addon.depends_on = dependency
+          (pure.extensions || []).each do |ext|
+            to_process << [ext, addon]
```
lslezak: Is it safe to change `to_process` while we are iterating over it?

src/lib/registration/addon.rb
```diff
@@ -125,23 +169,46 @@ def create_addon_with_deps(root)
...
+    # returns status of addon. Potential statuses are :registered, :selected, :auto_selected,
+    # :available and :none.
+    # @return [Symbol]
+    def status
+      return :registered if registered?
+      return :selected if selected?
+      return :auto_selected if auto_selected?
+      return :available if available?
+
+      :none
```
lslezak: Maybe `:unknown` would be a bit better here? An addon always has a state, in this case we just do not know which one.

I guess this can only happen if you pass some addon created from outside, which the code does not know. An then `:unknown` sounds good...

src/lib/registration/addon.rb
jreidinger: yes, if it is array, for Hashes and Sets exception is raised. ( tests cover that )

src/lib/registration/addon.rb
jreidinger: @lslezak well, I am thinking about good name. This state is in fact `:unavailable` - so not selected and also not available on scc. But I do not like it much.

lslezak: If it is not available in SCC then we do not know it and thus `:unknown` still makes sense. Probably not the best name, but IMHO still better than `:none`.

And this is just a theoretical case, normally we should never reach this pont. Maybe we should log a warning if it actually happens?

jreidinger: make sense

PR #336, opened 2017-10-05
Take users selection in AY profile into account when searching for base product

mchf:
fate#323450

do not merge before https://github.com/yast/yast-autoinstallation/pull/346

### DIFF
--- a/package/yast2-registration.spec
+++ b/package/yast2-registration.spec
@@ -17,7 +17,7 @@
 
 
 Name:           yast2-registration
-Version:        4.0.5
+Version:        4.0.6
 Release:        0
 
 BuildRoot:      %{_tmppath}/%{name}-%{version}-build
--- a/src/lib/registration/sw_mgmt.rb
+++ b/src/lib/registration/sw_mgmt.rb
@@ -45,7 +45,6 @@ module Registration
   Yast.import "Installation"
   Yast.import "PackageCallbacks"
   Yast.import "Popup"
-  Yast.import "AutoinstConfig"
 
   class SwMgmt
     include Yast
@@ -122,14 +121,16 @@ module Registration
 
       # use the selected product if a product has been already selected
       selected = product_selected? if Stage.initial
-      selected = !AutoinstConfig.selected_product.nil? if Mode.auto
 
       # during installation the products are :selected,
       # on a running system the products are :installed
       # during upgrade use the newer selected product (same as in installation)
       products = Pkg.ResolvableProperties("", :product, "").find_all do |p|
         if Stage.initial && Mode.auto
-          p["short_name"] == AutoinstConfig.selected_product.short_name
+          Yast.import "AutoinstConfig"
+          # note: AutoinstConfig.selected_product should never be nil when
+          # AY let it pass here
+          p["name"] == AutoinstConfig.selected_product.name
         elsif Stage.initial && !Mode.update
           # during installation the type is not valid yet yet
           # (the base product is determined by /etc/products.d/baseproduct symlink)
--- a/src/lib/registration/sw_mgmt.rb
+++ b/src/lib/registration/sw_mgmt.rb
@@ -45,6 +45,7 @@ module Registration
   Yast.import "Installation"
   Yast.import "PackageCallbacks"
   Yast.import "Popup"
+  Yast.import "AutoinstConfig"
 
   class SwMgmt
     include Yast
@@ -121,12 +122,15 @@ module Registration
 
       # use the selected product if a product has been already selected
       selected = product_selected? if Stage.initial
+      selected = !AutoinstConfig.selected_product.nil? if Mode.auto
 
       # during installation the products are :selected,
       # on a running system the products are :installed
       # during upgrade use the newer selected product (same as in installation)
       products = Pkg.ResolvableProperties("", :product, "").find_all do |p|
-        if Stage.initial && !Mode.update
+        if Stage.initial && Mode.auto
+          p["short_name"] == AutoinstConfig.selected_product.short_name
+        elsif Stage.initial && !Mode.update
           # during installation the type is not valid yet yet
           # (the base product is determined by /etc/products.d/baseproduct symlink)
           # use the selected product or the product from the first repository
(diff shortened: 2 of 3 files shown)


src/lib/registration/sw_mgmt.rb
```diff
@@ -121,12 +122,15 @@ def self.find_base_product
...
       selected = product_selected? if Stage.initial
+      selected = AutoinstConfig.selected_product if Mode.auto
 
       # during installation the products are :selected,
       # on a running system the products are :installed
       # during upgrade use the newer selected product (same as in installation)
       products = Pkg.ResolvableProperties("", :product, "").find_all do |p|
-        if Stage.initial
+        if Stage.initial && Mode.auto
+          p["short_name"] == selected
```
lslezak: Is it really `short_name`? That is a name suitable for displaying in the UI, for product identification we should use the `name` attribute. The short name usually includes a SP level (e.g. `SLES12-SP3`) that means the profile would not work with the SP+1 version and would need to be manually adjusted.

src/lib/registration/sw_mgmt.rb
```diff
@@ -121,12 +122,15 @@ def self.find_base_product
 
       # use the selected product if a product has been already selected
       selected = product_selected? if Stage.initial
+      selected = AutoinstConfig.selected_product if Mode.auto
```
lslezak: Please use a different variable, `selected` is a Boolean containing a flag whether a product has been already selected (see `self.product_selected?` below). Using a different type (String) in the AY mode is fragile and makes the code hard to read.

Use e.g. `ay_product_name` or something like that.

```ruby
if Mode.auto
  Yast.import "AutoinstConfig"
  ay_product_name = AutoinstConfig.selected_product
end

File: package/yast2-registration.spec [hunk 3]
```diff
@@ -47,6 +47,7 @@ Requires:       yast2-add-on >= 3.1.8
 # packager/product_patterns.rb
 Requires:       yast2-packager >= 3.1.95
 Requires:       yast2-update >= 3.1.36
+Requires:	autoyast2-installation
```
lslezak: We should avoid this dependency, this will pull whole AY stack into the installed system where is does not makes sense. Autoyast is listed as a [skelcd-control dependency](https://github.com/yast/skelcd-control-leanos/blob/master/package/skelcd-control-leanos.spec#L48) so it will be included in inst-sys anyway.

src/lib/registration/sw_mgmt.rb
```diff
@@ -45,6 +45,7 @@ module Registration
   Yast.import "Installation"
   Yast.import "PackageCallbacks"
   Yast.import "Popup"
+  Yast.import "AutoinstConfig"
```
lslezak: Move the module import to the code below so we do not need it in installed system.

src/lib/registration/sw_mgmt.rb
```diff
@@ -121,12 +121,16 @@ def self.find_base_product
 
       # use the selected product if a product has been already selected
       selected = product_selected? if Stage.initial
+      selected = !AutoinstConfig.selected_product.nil? if Mode.auto
```
lslezak: This line is not need, the call has been moved below.

package/yast2-registration.spec
```diff
@@ -58,6 +58,7 @@ BuildRequires:  yast2-slp >= 3.1.9
 # packager/product_patterns.rb
 BuildRequires:  yast2-packager >= 3.1.95
 BuildRequires:  yast2-update >= 3.1.36
+BuildRequires:	autoyast2-installation
```
lslezak: With 
```ruby
stub_const("Yast::AutoinstConfig", double(selected_product: ...))
```
we can avoid also this build dependency.

src/lib/registration/sw_mgmt.rb
```diff
@@ -121,12 +121,16 @@ def self.find_base_product
...
 
       # during installation the products are :selected,
       # on a running system the products are :installed
       # during upgrade use the newer selected product (same as in installation)
       products = Pkg.ResolvableProperties("", :product, "").find_all do |p|
-        if Stage.initial
+        if Stage.initial && Mode.auto
+          Yast.import "AutoinstConfig"
+          p["short_name"] == AutoinstConfig.selected_product.short_name
```
lslezak: Sorry, I have to repeat again:

Is it really `short_name`? That is a name suitable for displaying in the UI, for product identification we should use the `name` attribute. The short name usually includes a SP level (e.g. `SLES12-SP3`) that means the profile would not work with the SP+1 version and would need to be manually adjusted.

Or is there any reason for `short_name`?

lslezak:
BTW there some conflicts to resolve...

src/lib/registration/sw_mgmt.rb
mchf: no, I've just used that during testing ... I'll clean that.

package/yast2-registration.spec
mchf: hm, yes. I'll do so.

mchf:
I know there are conflicts. I'll solve them before merging. Currently this PR depends on another one (added to description).

src/lib/registration/sw_mgmt.rb
mchf: btw ```selected_product``` is a base product available on media. In fact it is ```Y2Packager::Product``` so it doesn't depend much if the selection is done using ```name``` or ```short_name```

src/lib/registration/sw_mgmt.rb
```diff
@@ -126,7 +126,12 @@ def self.find_base_product
       # on a running system the products are :installed
       # during upgrade use the newer selected product (same as in installation)
       products = Pkg.ResolvableProperties("", :product, "").find_all do |p|
-        if Stage.initial
+        if Stage.initial && Mode.auto
+          Yast.import "AutoinstConfig"
+          # note: AutoinstConfig.selected_product should never be nil when
+          # AY let it pass here
+          p["name"] == AutoinstConfig.selected_product.name
```
lslezak: Is the `AutoinstConfig.selected_product.name` value always set? What if it is missing in the profile? I'm afraid of the `nil` value here...

mchf: If no product can be recognized from the profile then AY currently reports error and aborts installation. If this behavior is changed this needs to be updated too.

lslezak: I see, so it is checked earlier. OK.

mchf:
autoyast changes finally merged, so merging even this one. Thanks for the review.

PR #338, opened 2017-10-11
Fix registration

jreidinger:


src/lib/registration/addon.rb
```diff
@@ -80,6 +80,27 @@ def registered_not_installed
...
+      def registration_order(list)
+        to_process = list.dup
+        result = []
+
+        loop do
+          break if to_process.empty?
+          next_addon = to_process.find do |addon|
+            addon.depends_on.nil? || !to_process.include?(addon.depends_on)
+          end
+          raise "circular dependencies found in addons #{to_process.inspect}" unless next_addon
```
lslezak: `Report.Error` would be probably more user friendly, but hopefully this never happens so let's make it simple...

test/addon_spec.rb
```diff
@@ -78,6 +78,19 @@
     end
   end
 
+  describe ".registration_order" do
+    it "returns in addons in order that allow registration" do
```
lslezak: `returns addons sorted in the registration order`

package/yast2-registration.changes
```diff
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Wed Oct 11 08:49:16 UTC 2017 - jreidinger@suse.com
+
+- Fix registration when there is auto selected extensions
```
lslezak: `Fixed registration of autoselected extensions`

src/lib/registration/addon.rb
jreidinger: Well, I hope we catch it quickly in openQA when this happen, so I hope not needed. Also Report is useless as result is that it won't work.

lslezak: The problem is that it might happen *after* the release, few months later... But anyway as this is really a corner case let's use this simple solution...

PR #339, opened 2017-10-11
Self update

jreidinger:


src/lib/registration/registration.rb
```diff
@@ -177,7 +178,14 @@ def migration_products(installed_products)
     # @see SwMgmt.remote_product
     # @see SUSE::Connect::Yast.list_installer_updates
     def get_updates_list(product = nil)
-      product ||= SwMgmt.base_product_to_register
+      if product.nil?
+        product = SwMgmt.base_product_to_register
+        id = Yast::ProductFeatures.GetStringFeature("globals", "self_update_id")
+        if !id.empty?
+          log.info "Using self update id from control file #{id.inspect}"
+          product["name"] = id
```
lslezak: I think we should check whether that specified product is available and use it if found. This would avoid problems when the products on the medium have different versions. That should not happen, but you never know...

BTW is it possible to add an unit test for this? Should not be that hard I guess...

jreidinger: we already discuss it yesterday. I am not fan of making self update id depending on product name. Maybe they want to change it to SLE or SLE-INSTALLER and then we do not have matching product on media.

Test make sense

lslezak: If we do not find any then we could fallback to this code and take random one. But if we find it we should prefer it to avoid problems with different versions.

src/lib/registration/registration.rb
```diff
@@ -167,24 +167,19 @@ def migration_products(installed_products)
...
-        if !id.empty?
-          log.info "Using self update id from control file #{id.inspect}"
-          product["name"] = id
-        end
+    def get_updates_list
+      product = SwMgmt.base_product_to_register
+      id = Yast::ProductFeatures.GetStringFeature("globals", "self_update_id")
+      if !id.empty?
+        log.info "Using self update id from control file #{id.inspect}"
+        product["name"] = id
```
lslezak: I'd add a comment saying that we expect all products on the medium to be in the same version so we can replace only the name attribute. This is an expectation which is not obvious from the code.

jreidinger: Yes, makes sense for me.

PR #341, opened 2017-10-26
Media based upgrade warnings, preselect recommended addons, better help text

lslezak:
# Summary

This pull request contains several improvements, I merged them into a single PR as they are quite simple. If you want to see the isolated changes then review each commit separately.

Note: I'll ask Ken for a feedback, maybe something will change a bit later, esp. the texts.

## Media Based Upgrade

Media based upgrade can be triggered by `upgrade=1` (or selecting the "Upgrade" item in the boot menu) and using `media_upgrade=1` boot option.

#### Unregistered System

When the system is not registered user have to use the DVD media for system upgrade and this message will be displayed:

media_upgrade

#### Registered System

If the system is registered we display some additional warning. The problem is that in this case we do not upgrade the registration status. The system is then in inconsistent state and we cannot provide support for this.

This is an experimental feature for experienced users who are aware of the consequences and are able to fix the possible problems manually.

media_upgrade_registered

## Preselect the Recommended Addons

SCC provides an extra flag indicating whether the addon should be selected by default. This solves the issue with the SLE15 module split and also addresses the old problem we had on ARM ([fate#320679](https://fate.suse.com/320679)) which we can remove now. See [bsc#1056413](https://bugzilla.suse.com/show_bug.cgi?id=1056413) for more details.

This feature is already implemented on the SCC side, currently the "Base system" module is marked as recommended.

The preselection behaves as if the modules were selected by the user. Unfortunately we cannot use the "auto selected" state as it would be removed by the dependency solver (nothing requires that module).

preselect_addons

## Improved Help Text

The help text has been improved to include the description of the check box states. This is especially important for the "auto selected" state which is specific to this dialog and is not used anywhere else.

The help text is properly displayed in both graphical and text mode.

##### Graphical Mode

addon_help_gui

##### Text Mode
addon_help_tui

## Test Update

I have updated the SLE15 addons fixture (`sle15_addons.yaml`) with the latest real data from the SCC to test the real current state.

### DIFF
--- a/src/lib/registration/ui/migration_repos_workflow.rb
+++ b/src/lib/registration/ui/migration_repos_workflow.rb
@@ -402,7 +402,11 @@ module Registration
         # media based upgrade requested by user
         if Yast::Linuxrc.InstallInf("MediaUpgrade") == "1"
           log.info "Skipping SCC upgrade, media based upgrade requested"
-          Yast::Popup.LongMessage(media_upgrade)
+          if Registration.is_registered?
+            Yast::Popup.LongMessageGeometry(media_upgrade(true), 60, 15)
+          else
+            Yast::Popup.LongMessage(media_upgrade(false))
+          end
           return :skip
         # the system is registered, continue with the SCC/SMT based upgrade
         elsif Registration.is_registered?
@@ -440,16 +444,28 @@ module Registration
       end
 
       # Informative message
+      # @param registered [Boolean] is the system registered?
       # @return [String] translated message
-      def media_upgrade
-        # TRANSLATORS: Media based upgrade requested by user (1/2)
+      def media_upgrade(registered)
+        # TRANSLATORS: Media based upgrade requested by user (1/3)
         #   User requested media based upgrade which does not use SCC/SMT
         #   but the downloaded media (physical DVD or shared repo on a local server).
-        _("<h2>Media Based Upgrade</h2><p>The media based upgrade is requested. " \
+        ret = _("<h2>Media Based Upgrade</h2><p>The media based upgrade is requested. " \
           "In this mode YaST will not contact the registration server to obtain " \
           "the new software repositories required for migration.</p>") +
-          # TRANSLATORS: Media based upgrade requested by user (2/2)
+          # TRANSLATORS: Media based upgrade requested by user (2/3)
           _("<p>Please add the installation media manually in the next step.</p>")
+
+        return ret unless registered
+
+        # TRANSLATORS: a warning message, upgrading the registered systems
+        #   using media is not supported
+        ret + _("<h2>Warning!</h2><p><b>The media based upgrade for registered " \
+          "systems is not supported!<b></p>") +
+          _("<p>If you upgrade the system using media the registration status " \
+            "will not be upgraded and the system will be still registered " \
+            "using the previous product. The packages from the registration " \
+            "repositories can conflict with the new packages.</p>")
       end
     end
   end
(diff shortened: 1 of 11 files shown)


src/lib/registration/ui/migration_repos_workflow.rb
```diff
@@ -440,16 +444,28 @@ def unregistered_message
...
           _("<p>Please add the installation media manually in the next step.</p>")
+
+        return ret unless registered
+
+        # TRANSLATORS: a warning message, upgrading the registered systems
+        #   using media is not supported
+        ret + _("<h2>Warning!</h2><p><b>The media based upgrade for registered " \
+          "systems is not supported!<b></p>") +
+          _("<p>If you upgrade the system using media the registration status " \
+            "will not be upgraded and the system will be still registered " \
```
imobachgs: np: I would say that "will not be updated" (the status). But I guess Ken will review the texts anyway...

test/addon_selection_dialog_test.rb
```diff
@@ -70,6 +64,52 @@
...
+    it "preselects the recommended addons" do
+      addon = addon_generator("recommended" => true)
+
+      expect(Yast::UI).to receive(:UserInput).and_return(:next)
+      registration = double(activated_products: [], get_addon_list: [addon])
+      expect(Registration::Addon).to receive(:selected).and_return([]).at_least(:once)
+
+      expect(Yast::Wizard).to receive(:SetContents) do |_title, content, _help, _back, _next|
+        # check the displayed content
+        term = content.nested_find do |t|
```
imobachgs: oh, I did not know about this `nested_find` method 👍

test/addon_selection_dialog_test.rb
```diff
@@ -70,6 +64,52 @@
       expect(wrapped_addon.selected?).to eq true
     end
 
+    it "preselects the recommended addons" do
+      addon = addon_generator("recommended" => true)
+
+      expect(Yast::UI).to receive(:UserInput).and_return(:next)
```
imobachgs: These three lines are pretty much the same in both tests. I would consider then as preconditions so maybe you could move them to a `before` block tunning them using `let` calls.

test/addon_selection_dialog_test.rb
```diff
@@ -70,6 +64,52 @@
...
+        # the check box is checked
+        expect(term.params[1]).to include("checkbox-on.png")
+        expect(term.params[1]).to_not include("checkbox-off.png")
+      end
+
+      subject.run(registration)
+    end
+
+    it "does not preselect the recommended addons if something is already selected" do
+      addon = addon_generator("recommended" => true)
```
imobachgs: why not use a `let` instead of defining into the test body? Same for `fake_addon`.

test/addon_selection_dialog_test.rb
```diff
@@ -70,6 +64,52 @@
...
+      # just to have an unknown, but "selected" addon for the Addon.selected call
+      fake_addon = addon_generator
+
+      expect(Yast::UI).to receive(:UserInput).and_return(:next)
+      registration = double(activated_products: [], get_addon_list: [addon])
+      expect(Registration::Addon).to receive(:selected).and_return([fake_addon]).at_least(:once)
+
+      # check the displayed content
+      expect(Yast::Wizard).to receive(:SetContents) do |_title, content, _help, _back, _next|
```
imobachgs: np: what about extracting this logic to a method? It is duplicated and it is not that rivial. Anyway, I do not have a strong opinion about it.

PR #342, opened 2017-10-26
fixed detection of base product during upgrade (bsc#1064191)

jsrain:


src/lib/registration/sw_mgmt.rb
```diff
@@ -136,6 +137,9 @@ def self.find_base_product
           # (the base product is determined by /etc/products.d/baseproduct symlink)
           # use the selected product or the product from the first repository
           selected ? p["status"] == :selected : p["source"] == 0
+        elsif Stage.initial
+          # during upgrade it depends on whether target is already initialized
```
lslezak: I'd add a comment to describe that specific case
```ruby
# use the product from the medium for the self-update step
```

PR #343, opened 2017-10-27
Small fix for displaying the AY summary (in installed system) (bsc#1065432)

lslezak:
- The original issue was caused by not moving the relevant `include` call, the fix itself was trivial, see the first commit (6afb30a)
- Introduced in Ruby 2.4 fix (https://github.com/yast/yast-registration/pull/332) so SLE12 is not affected
- I have refactored the code and added an unit test for that
- Also tested manually
- Related bug report: [bsc#1065432](https://bugzilla.suse.com/show_bug.cgi?id=1065432)

### DIFF
--- a/test/erb_renderer_spec.rb
+++ b/test/erb_renderer_spec.rb
@@ -1,0 +1,11 @@
+require_relative "spec_helper"
+
+require "registration/erb_renderer.rb"
+
+describe Registration::ErbRenderer do
+  subject { Registration::ErbRenderer.new(::Registration::Storage::Config.instance) }
+
+  it "renders an ERB file" do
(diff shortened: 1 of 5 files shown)


test/erb_renderer_spec.rb
```diff
@@ -0,0 +1,11 @@
+require_relative "spec_helper"
+
+require "registration/erb_renderer.rb"
+
+describe Registration::ErbRenderer do
+  subject { Registration::ErbRenderer.new(::Registration::Storage::Config.instance) }
+
+  it "renders an ERB file" do
+    expect(subject.render_erb_template("autoyast_summary.erb")).to be_a(String)
```
imobachgs: np: what about partially matching the content?

PR #345, opened 2017-11-08
[Review] Request from 'schubi2' @ 'yast/yast-registration/review_171108_warning_text_changed'

schubi2:
Please review the following changes:
  * 682fea8 changed warning text if user has skipped registration

### DIFF
--- a/src/lib/registration/ui/base_system_registration_dialog.rb
+++ b/src/lib/registration/ui/base_system_registration_dialog.rb
@@ -362,7 +362,7 @@ module Registration
         # TRANSLATORS:
         # %{media_name} is the media name (e.g. SLE-15-Packages),
         # %{download_url} is an URL link (e.g. https://download.suse.com)
-        warning = _("Without registration, update-channels will not be\n" \
+        warning = _("Without registration, update channels will not be\n" \
           "configured. This disables updates and security fixes.\n\n" \
           "A full system can be installed using the\n" \
           "%{media_name} media from %{download_url}.\n" \
--- a/src/lib/registration/ui/base_system_registration_dialog.rb
+++ b/src/lib/registration/ui/base_system_registration_dialog.rb
@@ -359,12 +359,16 @@ module Registration
       # @return [Boolean] true when skipping has been confirmed
       def show_skipping_warning
         # Popup question: confirm skipping the registration
+        // TRANSLATORS:
+        // %{media_name} is the media name (e.g. SLE-15-Packages),
+        // %{download_url} is an URL link (e.g. https://download.suse.com)
         warning = _("Without registration, update-channels will not be\n" \
             "configured. This disables updates and security fixes.\n\n" \
             "A full system can be installed using the\n" \
-            "%{media_name} media from https://download.suse.com.\n" \
+            "%{media_name} media from %{download_url}.\n" \
             "Without these media only a minimum system is available\n" \
-            "in this installation.") % {media_name: "SLE-15-Packages"}
+            "in this installation.")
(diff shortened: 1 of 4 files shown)


src/lib/registration/ui/base_system_registration_dialog.rb
```diff
@@ -359,10 +359,12 @@ def help_text
       # @return [Boolean] true when skipping has been confirmed
       def show_skipping_warning
         # Popup question: confirm skipping the registration
-        warning = _("If you do not register your system we will not be able\n" \
-            "to grant you access to the update repositories.\n\n" \
-            "You can register after the installation or visit our\n" \
-            "Customer Center for online registration.")
+        warning = _("Without registration, update-channels will not be\n" \
```
mvidner: "update channels" without the hyphen

lslezak: I'm not sure about the wording here... @mvidner ?

src/lib/registration/ui/base_system_registration_dialog.rb
```diff
@@ -359,10 +359,12 @@ def help_text
...
+            "configured. This disables updates and security fixes.\n\n" \
+            "A full system can be installed using the\n" \
+            "SLE-15-Packages media from download.suse.com.\n" \
```
mvidner: "media" is a plural. The singular is "medium".
Is the medium name correct?
How stable is it? Should we insert it with `format`?

src/lib/registration/ui/base_system_registration_dialog.rb
```diff
@@ -359,10 +359,12 @@ def help_text
...
+            "Without this media only a minimum system is available\n" \
```
mvidner: "medium"

schubi2: Hm, I have asked a native speaker to check these strings and she has approved it.

src/lib/registration/ui/base_system_registration_dialog.rb
lslezak: The medium name is correct. However I would not hard code it here but rather use a string interpolation `%{media_name}` placeholder as it still can be changed. The same with the site name.

That allows us to update the names without breaking the translations.

BTW I'd probably prefer `https://download.suse.com` value so it's clear what does it mean.

src/lib/registration/ui/base_system_registration_dialog.rb
lslezak: It's a bit complicated. It's actually a set of 3 media, but in reality just the DVD1 is enough (DVD2 = sources, DVD3 = debug packages IIRC).

mvidner: OK, if we are talking about several pieces of storage media, then "these media".

src/lib/registration/ui/base_system_registration_dialog.rb
```diff
@@ -359,10 +359,16 @@ def help_text
...
       def show_skipping_warning
         # Popup question: confirm skipping the registration
-        warning = _("If you do not register your system we will not be able\n" \
-            "to grant you access to the update repositories.\n\n" \
-            "You can register after the installation or visit our\n" \
-            "Customer Center for online registration.")
+        # TRANSLATORS:
+        # %{media_name} is the media name (e.g. SLE-15-Packages),
+        # %{download_url} is an URL link (e.g. https://download.suse.com)
+        warning = _("Without registration, update-channels will not be\n" \
```
mvidner: "update channels" without the hyphen

PR #348, opened 2017-12-18
Support for reading reg code from install.inf

mchf:
a way how to enable registration in AY integration testsuite (e.g. see comments in https://github.com/yast/aytests-tests/pull/102)

### DIFF
--- a/package/yast2-registration.changes
+++ b/package/yast2-registration.changes
@@ -4,6 +4,72 @@ Thu Dec 14 08:45:40 UTC 2017 - mfilka@suse.com
 - added support for reading registration code from install.inf
 
 -------------------------------------------------------------------
+Wed Nov 22 09:52:16 UTC 2017 - lslezak@suse.cz
+
+- AutoYaST fixes (bsc#1065438):
+  - fixed addon registration order
+  - automatically register the dependant addons
+- Preselect also the recommended beta addons, display them by
+  default even when the beta filter is active
+  (related to bsc#1056413)
+- 4.0.13
+
+-------------------------------------------------------------------
+Wed Nov  8 14:36:05 CET 2017 - schubi@suse.de
+
+- Changed warning text if user has skipped registration.
+  (bnc#1055023)
+- 4.0.12
+
+-------------------------------------------------------------------
+Fri Oct 27 12:44:58 UTC 2017 - lslezak@suse.cz
+
+- Do not preselect the default addons when some addon is already
+  registered (avoid selecting again when going back after
+  registration) (related to bsc#1056413)
+- 4.0.11
+
+-------------------------------------------------------------------
+Fri Oct 27 11:18:10 UTC 2017 - lslezak@suse.cz
+
+- Do not crash when displaying the AutoYaST summary in installed
+  system (bsc#1065432)
+- 4.0.10
+
+-------------------------------------------------------------------
+Thu Oct 26 12:42:18 UTC 2017 - lslezak@suse.cz
+
+- Use the correct install.inf key for media base upgrade (related
+  to fate#323163)
+- Warn when doing media based upgrade for a registered system
+- Preselect the recommended modules and extensions (bsc#1056413)
+- Describe the icons in the modules and extensions dialog help
+- 4.0.9
+
+-------------------------------------------------------------------
+Thu Oct 26 08:35:42 UTC 2017 - jsrain@suse.cz
+
+- fixed detection of base product during upgrade (bsc#1064191)
+- 4.0.8
+
+-------------------------------------------------------------------
+Thu Oct 19 11:53:37 UTC 2017 - lslezak@suse.cz
+
+- Display an informative popup when upgrading an unregistered
+  system, preselect adding DVD add-ons to make upgrade using media
+  easier (fate#323163)
+- Support "media_upgrade=1" boot parameter for media based upgrade
+  even for registered systems (fate#323163)
+- 4.0.7
+
+-------------------------------------------------------------------
+Thu Oct 12 11:57:29 UTC 2017 - mfilka@suse.com
+
+- fate#323450
+  - when searching for base product check also AY profiles for
+    user's selection
+- 4.0.6 
+-------------------------------------------------------------------
 Wed Oct 11 11:51:31 UTC 2017 - jreidinger@suse.com
 
 - use self update id if defined instead of product name to make
--- a/package/yast2-registration.changes
+++ b/package/yast2-registration.changes
@@ -1,7 +1,9 @@
 -------------------------------------------------------------------
 Thu Dec 14 08:45:40 UTC 2017 - mfilka@suse.com
 
-- added support for reading registration code from install.inf
+- fate#323450
+  - added support for reading registration code from install.inf
+- 4.0.14
 
 -------------------------------------------------------------------
 Wed Nov 22 09:52:16 UTC 2017 - lslezak@suse.cz
--- a/package/yast2-registration.changes
+++ b/package/yast2-registration.changes
@@ -1,4 +1,9 @@
 -------------------------------------------------------------------
+Thu Dec 14 08:45:40 UTC 2017 - mfilka@suse.com
+
+- added support for reading registration code from install.inf
+
+-------------------------------------------------------------------
 Wed Oct 11 11:51:31 UTC 2017 - jreidinger@suse.com
 
 - use self update id if defined instead of product name to make
--- a/package/yast2-registration.spec
+++ b/package/yast2-registration.spec
@@ -17,7 +17,7 @@
 
 
 Name:           yast2-registration
-Version:        4.0.5
+Version:        4.0.13
 Release:        0
 
 BuildRoot:      %{_tmppath}/%{name}-%{version}-build
--- a/package/yast2-registration.spec
+++ b/package/yast2-registration.spec
@@ -58,7 +58,6 @@ BuildRequires:  yast2-slp >= 3.1.9
 # packager/product_patterns.rb
 BuildRequires:  yast2-packager >= 3.1.95
 BuildRequires:  yast2-update >= 3.1.36
-BuildRequires:  autoyast2-installation
 
 BuildArch:      noarch
 
--- a/package/yast2-registration.spec
+++ b/package/yast2-registration.spec
@@ -58,6 +58,7 @@ BuildRequires:  yast2-slp >= 3.1.9
 # packager/product_patterns.rb
 BuildRequires:  yast2-packager >= 3.1.95
 BuildRequires:  yast2-update >= 3.1.36
+BuildRequires:  autoyast2-installation
 
 BuildArch:      noarch
 
--- a/package/yast2-registration.spec
+++ b/package/yast2-registration.spec
@@ -17,7 +17,7 @@
 
 
 Name:           yast2-registration
-Version:        4.0.13
+Version:        4.0.14
 Release:        0
 
 BuildRoot:      %{_tmppath}/%{name}-%{version}-build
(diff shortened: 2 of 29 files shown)


mvidner:
Well, we already have something similar: https://github.com/yast/yast-registration/pull/241 "Read registration codes from a USB stick". Wouldn't it be enough to use that?

mchf:
not in AY tests

src/lib/registration/storage.rb
```diff
@@ -37,7 +38,10 @@ class RegCodes < Struct.new(:reg_codes)
 
       def initialize
         self.reg_codes = if Stage.initial
-          reg_codes_from_usb_stick || {}
+          raw_reg_code = Yast::Linuxrc.InstallInf("reg_code")
+          reg_code_from_install_inf = raw_reg_code ? [raw_reg_code.split(":", 2)].to_h : {}
```
imobachgs: np: It would be nice to, perhaps, move these two lines to a separate method and document the format you expect.

src/lib/registration/storage.rb
```diff
@@ -131,14 +132,18 @@ def export
...
 
+        product = Yast::AutoinstConfig.selected_product.short_name
+
         @do_registration = settings.fetch("do_registration", false)
         @reg_server = settings["reg_server"] || ""
         @slp_discovery = settings.fetch("slp_discovery", false)
         @reg_server_cert = settings["reg_server_cert"] || ""
         @email = settings["email"] || ""
-        @reg_code = settings["reg_code"] || ""
+        @reg_code = settings["reg_code"] || RegCodes.instance.reg_codes[product] || ""
```
mvidner: I think it is wrong to put such logic to this layer. I think it should go higher. @lslezak can you advise please?

src/lib/registration/storage.rb
```diff
@@ -37,7 +38,7 @@ class RegCodes < Struct.new(:reg_codes)
 
       def initialize
         self.reg_codes = if Stage.initial
-          reg_codes_from_usb_stick || {}
+          reg_codes_from_usb_stick || reg_code_from_install_inf
```
mvidner: This one single line of code that is [not covered by tests](https://coveralls.io/builds/14865538/source?filename=src%2Flib%2Fregistration%2Fstorage.rb#L41)... contains a typo.

src/lib/registration/registration_codes_loader.rb
```diff
@@ -54,6 +54,18 @@ def reg_codes_from_usb_stick
       nil
     end
 
+    # Loades registration code from /etc/install.inf
+    #
+    # Expected format is reg_code: <product>:<registration_code>
+    # for example: reg_code: sle15:abcdefghijkl
+    #
+    # @return [Hash{String => String, nil] The key of the hash is product name
```
mvidner: The type annotation is missing a `}`, and it does matter where we put it.
For consistency with the other methods the return type should be a `Hash{String => String}, nil`, not a `Hash{String => String, nil}` so we should return `... : nil` and the caller should use `reg_codes_from_foo || reg_codes_from_bar || {}`.

package/yast2-registration.changes
```diff
@@ -1,3 +1,8 @@
+-------------------------------------------------------------------
+Thu Dec 14 08:45:40 UTC 2017 - mfilka@suse.com
+
+- added support for reading registration code from install.inf
```
mvidner: ETOOFEWDIGITS, can we add [FATE#323450](https://fate.suse.com/323450) "Allow installation of different products with one installation medium" to this?
Also, is it OK not to bump the version?

src/lib/registration/registration_codes_loader.rb
```diff
@@ -54,6 +54,18 @@ def reg_codes_from_usb_stick
...
+    #                                      and the value is registration code
+    def reg_codes_from_install_inf
+      raw_reg_code = Yast::Linuxrc.InstallInf("reg_code")
+      raw_reg_code.include?(":") ? [raw_reg_code.split(":", 2)].to_h : {}
```
mvidner:
```
NoMethodError: undefined method `include?' for nil:NilClass
```

src/lib/registration/registration_codes_loader.rb
```diff
@@ -54,6 +54,18 @@ def reg_codes_from_usb_stick
...
 
+    # Loades registration code from /etc/install.inf
+    #
+    # Expected format is reg_code: <product>:<registration_code>
+    # for example: reg_code: sle15:abcdefghijkl
+    #
+    # @return [Hash{String => String, nil] The key of the hash is product name
+    #                                      and the value is registration code
+    def reg_codes_from_install_inf
+      raw_reg_code = Yast::Linuxrc.InstallInf("reg_code")
```
mvidner: Yast.import "Linuxrc"

src/lib/registration/registration_codes_loader.rb
mchf: ok

src/lib/registration/registration_codes_loader.rb
mchf: ok

src/lib/registration/registration_codes_loader.rb
mchf: ok

src/lib/registration/storage.rb
mchf: ok

package/yast2-registration.changes
mchf: well it was originally developed due to internal requirements, but I will add both.

src/lib/registration/storage.rb
```diff
@@ -131,14 +131,18 @@ def export
       end
 
       def import(settings)
+        Yast.import "AutoinstConfig"
+
         reset
 
+        product = Yast::AutoinstConfig.selected_product.short_name
```
lslezak: Um, does this work also in AY config mode in installed system?

package/yast2-registration.spec
```diff
@@ -58,6 +58,7 @@ BuildRequires:  yast2-slp >= 3.1.9
 # packager/product_patterns.rb
 BuildRequires:  yast2-packager >= 3.1.95
 BuildRequires:  yast2-update >= 3.1.36
+BuildRequires:  autoyast2-installation
```
lslezak: The only thing we need from that package is `Yast::AutoinstConfig.selected_product` which we mock here anyway.

I'd rather mock that YaST module completely and avoid this unnecessary build dependency. See [this code](https://github.com/yast/yast-packager/blob/master/test/test_helper.rb#L21-L28) how it's done in the packager.

src/lib/registration/storage.rb
```diff
@@ -37,7 +37,7 @@ class RegCodes < Struct.new(:reg_codes)
 
       def initialize
         self.reg_codes = if Stage.initial
-          reg_codes_from_usb_stick || {}
+          reg_codes_from_usb_stick || reg_codes_from_install_inf || {}
```
lslezak: So you cannot use both features together... If it is OK then it should be at least documented somewhere.

src/lib/registration/storage.rb
lslezak: Probably the [import in the scc_auto](https://github.com/yast/yast-registration/blob/master/src/clients/scc_auto.rb#L125) client.

And it should be probably skipped in the AY config mode (as mentioned above).

lslezak:
Documentation note: reading the keys from USB is documented [here](https://github.com/yast/yast-registration/wiki/Loading-Registration-Codes-From-an-USB-Storage-(Flash-Drive-HDD)). We should also document this feature at the wiki.

package/yast2-registration.changes
```diff
@@ -1,3 +1,10 @@
+-------------------------------------------------------------------
+Thu Dec 14 08:45:40 UTC 2017 - mfilka@suse.com
+
+- fate#323450
```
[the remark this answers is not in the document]
mchf: If we talk about reading reg code from install.inf, it can be easily used by end user ... and I already do so when testing manual installations

package/yast2-registration.spec
mchf: ok, will try

src/lib/registration/registration_codes_loader.rb
```diff
@@ -54,6 +55,18 @@ def reg_codes_from_usb_stick
...
+    # Loades registration code from /etc/install.inf
+    #
+    # Expected format is reg_code: <product>:<registration_code>
+    # for example: reg_code: sle15:abcdefghijkl
+    #
+    # @return [Hash{String => String}, nil] The key of the hash is product name
+    #                                       and the value is registration code
+    def reg_codes_from_install_inf
+      raw_reg_code = Yast::Linuxrc.InstallInf("reg_code") || ""
+      raw_reg_code.include?(":") ? [raw_reg_code.split(":", 2)].to_h : nil
```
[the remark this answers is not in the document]
mchf: well ... I don't know ... what do you think?

src/lib/registration/storage.rb
mchf: yes. By intention (we have to set rules for possible collisions somehow ...)

src/lib/registration/storage.rb
mchf: why not? ```selected_product``` just tries to identify a base product according to the information in the profile.

src/lib/registration/registration_codes_loader.rb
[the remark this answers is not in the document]
lslezak: For the SLE15 installation with few basic free modules one reg. code is enough. So if there is no plan for supporting paid extensions then it's good enough. But I have no experience with AY integration tests or plans for the future so it's hard to say...

@imobachgs @schubi2 ?

src/lib/registration/storage.rb
lslezak: And when the profile does not define the product (yet)?

src/lib/registration/registration_codes_loader.rb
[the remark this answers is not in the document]
schubi2: We have currently no plan regarding AY test and registration. So please go on and we will try to adapt the tests to your changes.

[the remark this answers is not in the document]
mchf: just for record: my position when implementing that was, that it was implemented mainly for internal use where one reg code is currently enough. Extending to multiple codes is easy enough (e.g. using comma separated list)

mchf:
@lslezak: Added [documentation](https://github.com/yast/yast-registration/wiki/Loading-Registration-Codes-From-the--etc-install.inf)

src/lib/registration/registration_codes_loader.rb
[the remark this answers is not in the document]
lslezak: OK, I'm fine with that, I just wanted to be sure we do no need to change it in the foreseeable future...

package/yast2-registration.spec
```diff
@@ -58,7 +58,6 @@ BuildRequires:  yast2-slp >= 3.1.9
 # packager/product_patterns.rb
 BuildRequires:  yast2-packager >= 3.1.95
 BuildRequires:  yast2-update >= 3.1.36
-BuildRequires:  autoyast2-installation
```
mvidner: Great, thanks!

mchf:
Thanks all for the reviews.

src/lib/registration/storage.rb
mvidner: Then we get a P1: https://bugzilla.suse.com/show_bug.cgi?id=1075723

mvidner: I mean, this method is also called when AY is not used at all.

mvidner: @mchf UGH, does this mean that this code was supposed to be moved to scc_auto abut we forgot to delete it here??

mchf: ups. will check

mvidner: Yes.

PR #351, opened 2018-01-16
Adjust the base product selection (related to bsc#1071742)

lslezak:
Ensure that the base product selection is not influenced by a possibly changed repository priority, use the `system-installation()` provides instead of the hardcoded repository `0`, just like the other code in the installer.

Fix for bug [bsc#1071742](https://bugzilla.suse.com/show_bug.cgi?id=1071742) might lower the priority of the initial installation repository and then the base product is selected from the `Packages` DVD instead from the `Installer` DVD (with ID 0) and the code then does not work correctly.

- Skips the products which are not marked by the `system-installation()` provides as a base product (only in installation, in an installed system these provides might not be available)
- Tested manually in the latest build
- 4.0.17

### DIFF
--- a/src/lib/registration/sw_mgmt.rb
+++ b/src/lib/registration/sw_mgmt.rb
@@ -117,6 +117,8 @@ module Registration
     end
 
     def self.find_base_product
+      # FIXME: refactor the code to use Y2Packager::Product
+
       # just for debugging:
       return FAKE_BASE_PRODUCT if ENV["FAKE_BASE_PRODUCT"]
 
@@ -151,10 +153,10 @@ module Registration
 
     # Evaluate the product if it is a base product depending on the current
     # system status.
-    # @param p Hash the product from pkg-bindings
+    # @param p [Hash] the product from pkg-bindings
     # @param selected [Boolean,nil] is any product selected?
     # @param installed [Boolean,nil] is any product istalled?
-    # @return [Boolean]
+    # @return [Boolean] true if it is a base product
     def self.evaluate_product(p, selected, installed)
       if Stage.initial && Mode.auto
         Yast.import "AutoinstFunctions"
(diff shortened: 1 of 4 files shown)


src/lib/registration/sw_mgmt.rb
```diff
@@ -154,6 +149,41 @@ def self.find_base_product
       products.first
     end
 
+    # Evaluate the product if it is a base product depending on the current
+    # system status.
+    # @param p Hash the product from pkg-bindings
```
imobachgs: Syntax should be `@param p [Hash] the product from pkg-bindings`.

src/lib/registration/sw_mgmt.rb
```diff
@@ -123,28 +124,22 @@ def self.find_base_product
       selected = product_selected? if Stage.initial
       installed = product_installed? if Stage.initial
 
+      # list of products defined by the "system-installation()" provides
+      system_products = Y2Packager::ProductReader.installation_package_mapping.keys
```
imobachgs: Suggestion: what about moving this logic as a class method to `Y2Packager::Product`? At some extent, `ProductReader` looks like an "implementation detail".

lslezak: Yes, originally I wanted to use `Y2Packager::Product`, but I found some problems with that:

- It does not support all attributes required by registration, e.g. `register_target`
- The registration code expects a `Hash`, the `["name"]` code would have to be changed to `.name` at *very lot* places (or we would need to add a `.to_h` conversion)
- At the end the product is converted to `SUSE::Connect::Remote::Product` before sending to SCC, maybe we could convert and use it earlier ?

That's not trivial so we have to postpone it, I have added at least a `FIXME` note...

PR #352, opened 2018-01-16
Try base product registration code first

mvidner:
Card: https://trello.com/c/K5b1qBj7
Quoting [bsc#1075551](https://bugzilla.suse.com/show_bug.cgi?id=1075551):
> When registering SLES4SAP, the same registration code as for the base product needs to be re-entered for the HA extension, which is part of the subscription.
>
>The installer should assure that for these modules registration code is not asked again.
>
> The idea is: Try to register non-free extensions with the same product as base system and only ask for those for which registration does not succeed.

Now, the twist in the plot is that the above problem is *hypothetical*, because until now, SLES4SAP is using a *free* variant of the HA extension and the problem does not occur.

I have fixed it anyway. At least I think so, I could not test the real thing. I tested the SLES + HA scenario, which uses a *paid* HA which is *not* included with SLES, and the code behaves in a way that I believe will work out.

### DIFF
--- a/src/lib/registration/connect_helpers.rb
+++ b/src/lib/registration/connect_helpers.rb
@@ -103,7 +103,8 @@ module Registration
           check_smt_api(error_msg)
           report_error(message_prefix + _("Connection to registration server failed."), error_msg)
         when 422
-          if silent_reg_code_mismatch && e.message =~ /does not include the requested product/
+          if silent_reg_code_mismatch && e.response.body["error"] =~
+              /does not include the requested product/
             log.info "Reg code does not work for this product, that's OK"
           else
             # Error popup
--- a/src/lib/registration/connect_helpers.rb
+++ b/src/lib/registration/connect_helpers.rb
@@ -128,39 +128,11 @@ module Registration
         false
       rescue OpenSSL::SSL::SSLError => e
         log.error "OpenSSL error: #{e}"
-
-        cert = Storage::SSLErrors.instance.ssl_failed_cert
-        error_code = Storage::SSLErrors.instance.ssl_error_code
-        expected_cert_type = Storage::Config.instance.reg_server_cert_fingerprint_type
-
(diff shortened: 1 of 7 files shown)


mvidner:
As of now, for the SLES + HA scenario, there is an error popup before we ask for the HA regcode. I will fix that.

src/lib/registration/connect_helpers.rb
```diff
@@ -96,8 +103,12 @@ def self.catch_registration_errors(message_prefix: "", show_update_hint: false,
           check_smt_api(error_msg)
           report_error(message_prefix + _("Connection to registration server failed."), error_msg)
         when 422
-          # Error popup
-          report_error(message_prefix + _("Connection to registration server failed."), error_msg)
+          if silent_reg_code_mismatch && e.message =~ /does not include the requested product/
```
lslezak: Be careful, the `message` text is a *translated* error message, this won't work in non-English locale.

See https://github.com/SUSE/connect/blob/master/lib/suse/connect/errors.rb#L35

We should use `e.response.body["error"]` instead.

mvidner:
(Code climate does not like `handle_ssl_error`, saying it is too complex. Yes, but I have just factored it out of a *larger* method.)

mvidner:
@lslezak Is there a related piece of documentation that I should update? AFAIK not.

lslezak:
To the documentation: there are two wiki pages:

- [Loading Registration Codes From an USB Storage (Flash Drive HDD)](https://github.com/yast/yast-registration/wiki/Loading-Registration-Codes-From-an-USB-Storage-(Flash-Drive-HDD))
- [Loading Registration Code From /etc/install.inf](https://github.com/yast/yast-registration/wiki/Loading-Registration-Codes-From-the--etc-install.inf)

You can add a note to both of them or create a new page and link it from the documents.

PR #353, opened 2018-01-18
SLE15 Offline Migration (fate#323163)

lslezak:
- See [FATE#323163](https://fate.suse.com/323163) for more details
- Let's ignore the CodeClimate complains, we need this for Beta6
- Also the unit tests will be added later
- Tested manually in an SLES12-SP3 -> SLES15 upgrade

## TODO

- [x] Depens on an SCC fix for ~~[bsc#1076513](https://bugzilla.suse.com/show_bug.cgi?id=1076513)~~ [FIXED]
- [x] Depens on an SCC fix for ~~[bsc#1076611](https://bugzilla.suse.com/show_bug.cgi?id=1076611)~~ [FIXED]
- [x] The product reading needs [this patch](https://github.com/yast/yast-yast2/pull/672) and [this another patch](https://github.com/yast/yast-packager/pull/311) [MERGED]

## Testing

If something goes wrong during the migration the old system might become broken (unregistered). In that case boot the old system and use commands
- `SUSEConnect --status-text` to check the registration status
- If the system is not registered, but was properly registered before then you can use `SUSEConnect --rollback` command to restore the previous registration. If it fails use `SUSEConnect --cleanup` and do a full registration again.

### DIFF
--- a/src/lib/registration/clients/inst_migration_repos.rb
+++ b/src/lib/registration/clients/inst_migration_repos.rb
@@ -39,10 +39,26 @@ module Registration
 
       # Pass the target directory to SUSEConnect
       def set_target_path
-        return if Yast::Installation.destdir == "/"
+        destdir = Yast::Installation.destdir
+        return if destdir.nil? || destdir == "/"
 
-        log.info("Setting SUSEConnect target directory: #{Yast::Installation.destdir}")
-        SUSE::Connect::System.filesystem_root = Yast::Installation.destdir
+        log.info("Setting SUSEConnect target directory: #{destdir}")
+        SUSE::Connect::System.filesystem_root = destdir
+
+        # copy the old config from the upgraded system to inst-sys
+        # to correctly work in the SMT case
+        # FIXME: this should not be needed, it should be possible to read
+        # the config from the /mnt directly...
+        target_path = SUSE::Connect::YaST::DEFAULT_CONFIG_FILE
+        source_path = File.join(destdir, target_path)
+
+        if File.exist?(source_path)
+          log.info("Copying #{source_path} -> #{target_path}")
+          ::FileUtils.cp(source_path, target_path)
+        elsif File.exist?(target_path)
+          log.info("Removing #{target_path}...")
+          ::FileUtils.rm(target_path)
+        end
       end
     end
   end
--- a/src/lib/registration/clients/inst_migration_repos.rb
+++ b/src/lib/registration/clients/inst_migration_repos.rb
@@ -39,8 +39,8 @@ module Registration
 
       # Pass the target directory to SUSEConnect
       def set_target_path
-        destdir = Yast::Installation.destdir
-        return if destdir.nil? || destdir == "/"
+        destdir = Yast::Installation.destdir || "/"
+        return if destdir == "/"
 
         log.info("Setting SUSEConnect target directory: #{destdir}")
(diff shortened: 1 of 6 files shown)


src/lib/registration/clients/inst_migration_repos.rb
```diff
@@ -39,10 +39,26 @@ def main
 
       # Pass the target directory to SUSEConnect
       def set_target_path
-        return if Yast::Installation.destdir == "/"
+        destdir = Yast::Installation.destdir
+        return if destdir.nil? || destdir == "/"
```
jreidinger: what about
```
destdir = Yast::Installation.destdir || "/"
return if destdir == "/"
```

src/lib/registration/registration.rb
```diff
@@ -167,6 +167,18 @@ def migration_products(installed_products)
       migrations
     end
 
+    def offline_migration_products(installed_products, target_base_product)
+      log.info "Offline migration for: #{target_base_product}."
+      migration_paths = []
+      ConnectHelpers.catch_registration_errors do
+        migration_paths = SUSE::Connect::YaST
+                          .system_offline_migrations(installed_products, target_base_product)
+      end
```
jreidinger: does it make sense to continue if there are not migration paths?

src/lib/registration/ui/migration_repos_workflow.rb
```diff
@@ -240,10 +247,22 @@ def merge_registered_addons
         products.concat(addons)
       end
 
-      # load migration products for the installed products from the registration server
+      # load migration products for the installed products from the registration server,
+      # loads online or offline migrations depending on the system state
       # @return [Symbol] workflow symbol (:next or :abort)
       def load_migration_products
-        log.info "Loading migration products from server"
+        if Yast::Stage.initial && Yast::Mode.update
```
jreidinger: what is reason to check also that Mode is update?

src/lib/registration/ui/migration_repos_workflow.rb
```diff
@@ -483,6 +533,23 @@ def media_upgrade(registered)
...
+
+      def upgraded_base_product
+        # temporarily run the update mode to let the solver select the product for upgrade
+        # (this will correctly handle possible product renames)
+        Yast::Pkg.PkgUpdateAll({})
+        product = Y2Packager::Product.selected_base
+
+        # restore the initial status, the package update will be turned on later again
+        Yast::Pkg.PkgReset
+        changed = Yast::Pkg.ResolvableProperties("", :package, "").select do |p|
```
jreidinger: Today we kill another instance of this problematic call and now we reintroducing it again. I worry it again kill memory requirements.

jreidinger: From those two use cases it looks like we need new call like "PackagesByStatus" that accept one or more statuses adn return package names for that.

lslezak:
We already have [Pkg.GetPackages](https://github.com/yast/yast-pkg-bindings/blob/ed7b6c815964c5ea9ef11b6da80d45cb1a4b2302/src/Package.cc#L1382-L1401) call.

src/lib/registration/registration.rb
lslezak: That's handled [elsewhere](https://github.com/yast/yast-registration/pull/353/files#diff-1be991fe1b75a62325e272cdd2ee099dR301).

PR #354, opened 2018-02-02
Support for the SLE11 migration

lslezak:
- Always copy the credentials file to inst-sys (reuse the existing code, removed some duplication)
- As the credentials are copied we do not need to handle the `/mnt` prefix in specific cases
- Manually tested with SLES11-SP3 and SLES11-SP4
- Also manually tested with SLES12-SP3 to avoid regressions

### DIFF
--- a/test/registration_spec.rb
+++ b/test/registration_spec.rb
@@ -271,7 +271,7 @@ describe Registration::Registration do
       expect(Registration::Registration.is_registered?).to eq(true)
     end
 
-    it "returns false if the global credentials file exists" do
+    it "returns false if the global credentials file does not exist" do
       expect(File).to receive(:exist?).with(SUSE::Connect::YaST::GLOBAL_CREDENTIALS_FILE)
         .and_return(false)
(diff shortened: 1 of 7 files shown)


test/registration_spec.rb
```diff
@@ -279,23 +265,16 @@
...
+      expect(File).to receive(:exist?).with(SUSE::Connect::YaST::GLOBAL_CREDENTIALS_FILE)
+        .and_return(true)
+      expect(Registration::Registration.is_registered?).to eq(true)
     end
 
-    it "does not add the prefix if not at upgrade" do
-      allow(Yast::Mode).to receive(:update).and_return(false)
-      allow(Yast::Stage).to receive(:initial).and_return(false)
-      expect(Yast::Installation).to_not receive(:destdir)
+    it "returns false if the global credentials file exists" do
```
gilsonsouza: "....if the global credential file **does not** exist", right?

lslezak: Oh, right, copy&paste error...

PR #356, opened 2018-02-19
Offline migration from SLE11 to SLE15 (fate#323395)

lslezak:
# Summary
- https://fate.suse.com/323395
- 4.0.22

## Notes

- Ignore the CodeClimate code complexity error, I do not want to do any unnecessary refactoring just before the RC phase...

## Included Fixes

- Display a hint for a failed migration from SLE11 (NCC<->SCC sync)
- Debugging: log the activated products for easier debugging (to see if the system was really registered and what is active)
- Ask for confirmation if the target product has been already registered (SCC fails when attempting to migrate an already migrated system)
- Remove the obsolete `/etc/zypp/credentials.d/NCCcredentials` file at the end, not needed anymore
- Skipping addon loading at offline upgrade (not needed)
- Fix for product rename (the internal identifier has been changed from `SUSE_SLES` in SLE11 to `SLES` in SLE12/15, also some SLE12 modules have been renamed or merged)

### NCC Hint

When the NCC registration (in SLE11) has not been synchronized yet to SCC then a hint is displayed how to speed up the synchronization:

sles11_offline_migration_missing_scc_sync

### Confirmation

Migrating an already migrated machine fails with an SCC error, to prevent from this display a confirmation popup:

sle11_already_migrated

### Product Rename

The product renames or merges are now handled correctly. Here the SDK and Toolchain Module have been merged into Development Tools (tested in upgrade from SLES12):

sle15_migration_product_renames

### DIFF
--- a/src/lib/registration/connect_helpers.rb
+++ b/src/lib/registration/connect_helpers.rb
@@ -328,8 +328,7 @@ module Registration
       msg = _("Check that this system is known to the registration server.")
 
       # probably missing NCC->SCC sync, display a hint unless SMT is used
-      if UrlHelpers.registration_url == SUSE::Connect::YaST::DEFAULT_URL ||
-          UrlHelpers.registration_url.nil?
+      if [nil, SUSE::Connect::YaST::DEFAULT_URL].include?(UrlHelpers.registration_url)
 
         msg += "\n\n"
         # TRANSLATORS: additional hint for an error message
--- a/src/lib/registration/connect_helpers.rb
+++ b/src/lib/registration/connect_helpers.rb
@@ -328,7 +328,9 @@ module Registration
       msg = _("Check that this system is known to the registration server.")
 
       # probably missing NCC->SCC sync, display a hint unless SMT is used
-      if UrlHelpers.registration_url == SUSE::Connect::YaST::DEFAULT_URL
+      if UrlHelpers.registration_url == SUSE::Connect::YaST::DEFAULT_URL ||
+          UrlHelpers.registration_url.nil?
+
         msg += "\n\n"
         # TRANSLATORS: additional hint for an error message
         msg += _("If you are upgrading from SLE11 make sure the SCC server\n" \
--- a/src/lib/registration/ui/migration_repos_workflow.rb
+++ b/src/lib/registration/ui/migration_repos_workflow.rb
@@ -13,6 +13,7 @@
 #
 
 require "yast"
+require "yast2/popup"
 
 require "registration/registration"
 require "registration/registration_ui"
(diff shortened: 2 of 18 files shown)


src/lib/registration/connect_helpers.rb
```diff
@@ -328,7 +328,9 @@ def self.add_update_hint(error_msg)
       msg = _("Check that this system is known to the registration server.")
 
       # probably missing NCC->SCC sync, display a hint unless SMT is used
-      if UrlHelpers.registration_url == SUSE::Connect::YaST::DEFAULT_URL
+      if UrlHelpers.registration_url == SUSE::Connect::YaST::DEFAULT_URL ||
+          UrlHelpers.registration_url.nil?
```
jreidinger: what about `[nil, SUSE::Connect::YaST::DEFAULT_URL].include?(UrlHelpers.registration_url)`

src/lib/registration/ui/migration_repos_workflow.rb
```diff
@@ -311,6 +319,33 @@ def load_migration_products_offline
...
 
+      # check if the target producte has been already activated, if yes
+      # require user confirmation for the migration (though it will very likely fail)
+      # @param base_product [Y2Packager::Product] the base product to which migrate to
+      # @param activations [Array<SUSE::Connect::Remote::Product>] already activated
+      #   products on the server
+      # @return [Boolean] true = continue with the migration, false = stop
+      def migration_confirmed?(base_product, activations)
+        # version without the release, e.g. "15-0" => "15"
+        base_product_version = base_product.version.split("-", 2).first
```
jreidinger: this is a bit confusing to use splitinto two  and then use just first one. Maybe use something like
```
base_product_version = base_product.version[/^[^-]*/]
# or less cryptic
base_product_version, _minor_version = base_product.version.split("-", 2)
```

src/lib/registration/ui/migration_selection_dialog.rb
```diff
@@ -181,7 +182,7 @@ def migration_details(idx)
 
         details = sorted_migrations[idx].map do |product|
           installed = installed_products.find do |installed_product|
-            installed_product["name"] == product.identifier
+            same_products?(installed_product, product)
```
jreidinger: NP: in object world this should be method of product like e.g. `=~`

src/lib/registration/ui/migration_selection_dialog.rb
```diff
@@ -192,11 +193,24 @@ def migration_details(idx)
...
+          migrated_product = sorted_migrations[idx].find do |product|
+            same_products?(installed_product, product)
+          end
+
+          details << "<li>" + product_summary(migrated_product, installed_product) + "</li>"
         end
 
         # TRANSLATORS: RichText header (details for the selected item)
-        "<h3>" + _("Migration Summary") + "</h3><ul>" + details.join + "</ul>"
+        det = "<h3>" + _("Migration Summary") + "</h3><ul>" + details.join + "</ul>"
```
jreidinger: NP: I prefer join with "\n" so you can easily log it.

src/lib/registration/ui/migration_selection_dialog.rb
```diff
@@ -192,11 +193,24 @@ def migration_details(idx)
...
+        puts det
```
jreidinger: probably forgotten debug line.

src/lib/registration/ui/migration_selection_dialog.rb
lslezak: In this cases the objects are of different classes, that makes it a bit more complicated... But yes you are right.

PR #357, opened 2018-03-05
Display more error details (related to bsc#1083503)

lslezak:
- Display more details when a service refresh fails
- For easier debugging to see where the error is (SMT, SCC or SCC proxy)
- We can easily detect non-YaST issues and reassign bugs quicker
- See the screen in openQA: https://openqa.suse.de/tests/1513698#step/scc_registration/8

### The old error was just
```
Refreshing service 'SUSE_Linux_Enterprise_Server_15_ppc64le' failed.
```
### The new error message will also contain details from libzypp
```
Refreshing service 'SUSE_Linux_Enterprise_Server_15_ppc64le' failed.
      
Details: Download (curl) error for 'http://all-489.1.proxy.scc.suse.de/access/services/1570/repo/repoi
ndex.xml?cookies=0&credentials=SUSE_Linux_Enterprise_Server_15_ppc64le':
Error code: HTTP response: 500
Error message: The requested URL returned error: 500 Internal Server Error
```

So the user will see which URL failed and in this case we can reassign the bug to the SCC Proxy maintainers.

### DIFF
--- a/src/lib/registration/connect_helpers.rb
+++ b/src/lib/registration/connect_helpers.rb
@@ -121,11 +121,11 @@ module Registration
         false
       rescue ::Registration::ServiceError => e
         log.error("Service error: #{e.message % e.service}")
-        Yast::Report.Error(_(e.message) % e.service)
+        report_pkg_error(_(e.message) % e.service)
         false
       rescue ::Registration::PkgError => e
         log.error("Pkg error: #{e.message}")
-        Yast::Report.Error(_(e.message))
+        report_pkg_error(_(e.message))
         false
       rescue OpenSSL::SSL::SSLError => e
         log.error "OpenSSL error: #{e}"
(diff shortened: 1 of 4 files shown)


src/lib/registration/connect_helpers.rb
```diff
@@ -121,11 +121,11 @@ def self.catch_registration_errors(message_prefix: "",
         false
       rescue ::Registration::ServiceError => e
         log.error("Service error: #{e.message % e.service}")
-        Yast::Report.Error(_(e.message) % e.service)
+        report_pkg_error(_(e.message) % e.service)
```
joseivanlopez: Only for curiosity: Can this message be correctly translated? The string is generated at runtime, so how to know what is the message?

lslezak: The message is defined here: https://github.com/yast/yast-registration/blob/6dbf98a245e104e3135cf1d87d305476e020dcf1/src/lib/registration/sw_mgmt.rb#L321

The `N_()` method only marks the string for translation, it returns the original (untranslated string). It's later translated via the `_()` call. See http://yastgithubio.readthedocs.io/en/latest/localization/#translating-constants

The reason is that the `log.error` above will log the original untranslated message. That's for easier debugging, reading errors logged in a non-Latin based language is tricky.

lslezak:
> Did you forget to increase the version in spec file?

@joseivanlopez good point, I thought I ran the rake task...

teclator:
@joseivanlopez good catch, so it now looks even better to me :)

PR #358, opened 2018-03-13
Do not crash if a product is not selected

teclator:
- https://bugzilla.suse.com/show_bug.cgi?id=1084820

The list of updates will return an empty list avoiding the crash and then an error will be reported during autoinit

errorshown


teclator:
@lslezak thnx for reviewing it.

PR #359, opened 2018-03-16
Use the first available product for installer update

teclator:
- [Trello Card](https://trello.com/c/syNlrXoP/1371-3-p1-1084820-handle-selfupdate-during-autoupgrade-correctly)

If there is a `self_update_id` defined in the control file, the product is not so important at the end as we will replace the product name and will reuse the `version` & `arch` just use the first available product as the template.

### DIFF
--- a/package/yast2-registration.changes
+++ b/package/yast2-registration.changes
@@ -1,4 +1,12 @@
 -------------------------------------------------------------------
+Fri Mar 16 13:22:35 UTC 2018 - knut.anderssen@suse.com
+
+- Just use the first product availble from the media When a self
+  update id is defined in the control file (bsc#1084820)
+- 4.0.25
+
+
+-------------------------------------------------------------------
 Tue Mar 13 09:37:14 UTC 2018 - knut.anderssen@suse.com
 
 - Prevent crashing when no product was defined in the profile
--- a/package/yast2-registration.changes
+++ b/package/yast2-registration.changes
@@ -1,10 +1,9 @@
 -------------------------------------------------------------------
 Fri Mar 16 13:22:35 UTC 2018 - knut.anderssen@suse.com
 
-- Just use the first product availble from the media When a self
+- Just use the first product available from the media When a self
   update id is defined in the control file (bsc#1084820)
 - 4.0.25
-
 
 -------------------------------------------------------------------
 Tue Mar 13 09:37:14 UTC 2018 - knut.anderssen@suse.com
--- a/test/registration_spec.rb
+++ b/test/registration_spec.rb
@@ -200,18 +200,15 @@ describe Registration::Registration do
   end
 
   describe "#get_updates_list" do
+    let(:self_update_id) { "SLES" }
     let(:base_product) { { "name" => "base" } }
-    let(:installer_update_base_product) { { "name" => "self_update_id" } }
+    let(:installer_update_base_product) { { "name" => self_update_id } }
     let(:remote_product) { { "name" => "base" } }
     let(:updates) { ["http://updates.suse.com/sles12/"] }
     let(:suse_connect) { double("suse_connect") }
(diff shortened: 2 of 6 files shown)


package/yast2-registration.changes
```diff
@@ -1,3 +1,11 @@
+-------------------------------------------------------------------
+Fri Mar 16 13:22:35 UTC 2018 - knut.anderssen@suse.com
+
+- Just use the first product availble from the media When a self
```
jreidinger: typo availble

test/registration_spec.rb
```diff
@@ -201,38 +201,48 @@
 
   describe "#get_updates_list" do
     let(:base_product) { { "name" => "base" } }
+    let(:installer_update_base_product) { { "name" => "self_update_id" } }
```
jreidinger: this should use variable and not string?

test/registration_spec.rb
```diff
@@ -201,38 +201,48 @@
...
+          .and_return(updates)
+        expect(subject.get_updates_list).to eq(updates)
+      end
     end
 
-    it "returns an empty list if no base product is selected" do
-      allow(Registration::SwMgmt).to receive(:base_product_to_register).and_return(nil)
-      expect(subject.get_updates_list).to eq([])
+    context "when the control file does not define a self_update_id" do
+      let(:self_update_id) { "" }
```
jreidinger: how this can work when above it is string and not var? does test really fail without this line?

test/registration_spec.rb
teclator: when it is returned means that the self_update_id variable is not empty if it is empty then the method should return nil not a hash

test/registration_spec.rb
teclator: because when it is empty the method returns nil

test/registration_spec.rb
```diff
@@ -201,38 +201,48 @@
...
     let(:remote_product) { { "name" => "base" } }
     let(:updates) { ["http://updates.suse.com/sles12/"] }
     let(:suse_connect) { double("suse_connect") }
+    let(:self_update_id) { "self_update_id" }
 
     before do
       allow(Registration::SwMgmt).to receive(:base_product_to_register).and_return(base_product)
+      allow(Yast::ProductFeatures).to receive(:GetStringFeature)
+        .with("globals", "self_update_id")
+        .and_return(self_update_id)
```
teclator: Here is where the self_update_id variable is used, so "self_update_id" for defined and "" for not defined

jreidinger: ah ok

test/registration_spec.rb
jreidinger: I really do not like this as I do not know if it is coincident name string or not. is posible to use here `"name" => self_update_id`? as it looks like connected values for me. And if it is not connect, please use different name.

test/registration_spec.rb
```diff
@@ -201,38 +201,48 @@
 
   describe "#get_updates_list" do
     let(:base_product) { { "name" => "base" } }
+    let(:installer_update_base_product) { { "name" => "self_update_id" } }
     let(:remote_product) { { "name" => "base" } }
     let(:updates) { ["http://updates.suse.com/sles12/"] }
     let(:suse_connect) { double("suse_connect") }
+    let(:self_update_id) { "self_update_id" }
```
lslezak: I'd suggest using a real value here, e.g. the current value is `SLES` (see https://github.com/yast/skelcd-control-leanos/blob/master/control/control.leanos.xml#L76).

Using `self_update_id` is confusing, it's like naming variable `variable`...

test/registration_spec.rb
lslezak: As I wrote below, using a real value like `"SLES"` here would make it clear.

teclator:
@lslezak thnx

PR #364, opened 2018-03-28
Use the new way for finding the product upgrade (bsc#1086259)

lslezak:
- Use the new ProductUpgrade class from https://github.com/yast/yast-yast2/pull/717


jreidinger:
looks good so far

PR #365, opened 2018-03-29
Use the new license API

teclator:
- https://trello.com/c/Ve9xuqt3/1377-8-licenses-help-customers-to-understand-that-they-are-accepting-another-license-for-another-product-no-you-know-what-just-make-i
- It depends on yast/yast-yast2#715

### DIFF
--- a/src/lib/registration/ui/addon_eula_dialog.rb
+++ b/src/lib/registration/ui/addon_eula_dialog.rb
@@ -3,6 +3,7 @@ require "yast"
 require "registration/eula_downloader"
 require "registration/eula_reader"
 require "registration/helpers"
+require "y2packager/license_store"
 
 module Registration
   module UI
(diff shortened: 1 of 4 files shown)


src/lib/registration/ui/addon_eula_dialog.rb
```diff
@@ -140,6 +145,15 @@ def accept_eula(addon)
       def display_optional_info(info_file)
         Yast::InstShowInfo.show_info_txt(info_file) if File.exist?(info_file)
       end
+
+      def find_license(addon, eula_reader)
+        license_file = eula_reader.licenses[Y2Packager::License::DEFAULT_LANG]
+        return false unless license_file
```
jreidinger: I think nil makes more sense here as result.

src/lib/registration/ui/addon_eula_dialog.rb
```diff
@@ -140,6 +145,15 @@ def accept_eula(addon)
...
+        content = Yast::SCR.Read(path(".target.string"), license_file)
+        return false unless content
```
jreidinger: and same here.

PR #366, opened 2018-04-06
Version bump

mchf:


package/yast2-registration.spec
```diff
@@ -17,7 +17,7 @@
 
 
 Name:           yast2-registration
-Version:        4.0.31
```
jreidinger: this looks really wrong!

mchf:
No reason for version bum here -> invalid PR

PR #368, opened 2018-04-10
Select release packages during SCC based migration

imobachgs:
* Fixes [bsc#1086818](https://bugzilla.suse.com/show_bug.cgi?id=1086818).
* Check the [Trello card](https://trello.com/c/h7Z1kKAU/) for further details.

### DIFF
--- a/src/lib/registration/ui/migration_repos_workflow.rb
+++ b/src/lib/registration/ui/migration_repos_workflow.rb
@@ -84,6 +84,7 @@ module Registration
           "register_migration_products"  => [->() { register_migration_products }, true],
           "activate_migration_repos"     => [->() { activate_migration_repos }, true],
           "select_migration_repos"       => ->() { select_migration_repos },
+          "select_products"              => ->() { select_products },
           "store_repos_state"            => ->() { store_repos_state }
         }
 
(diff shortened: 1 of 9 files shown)


src/lib/registration/ui/migration_repos_workflow.rb
```diff
@@ -466,6 +470,17 @@ def select_migration_repos
         UI::MigrationReposSelectionDialog.run
       end
 
+      # Select products for migration
+      #
+      # It causes the *-release packages to be installed (see bsc#1086818 for
+      # further details).
+      def select_products
+        selected_migration.map(&:identifier).each do |name|
```
lslezak: Could we somehow reuse the [SwMgmt.select_addon_products](https://github.com/yast/yast-registration/blob/4866c70e53c485a45711b53eb9f0ae7e5e1dc497/src/lib/registration/sw_mgmt.rb#L512) here?

Additionally it also selects the default patterns which we should not do in upgrade so we would need a new parameter.

But IMHO it works a bit better because it does not blindly expect the "product name" == "migration identifier", but uses the available product from that repository. Which means it should better handle the inconsistencies between the SCC identifier and the libzypp product. Although that's rather a corner case, normally they should match...

But I do not know if during the migration the used `Registration::Storage::Cache.instance.addon_services` value is set. Please check.

imobachgs: That was my first try, but during migration `addon_services` is always empty. But I will try to find out why `addon_services` is empty.

imobachgs:
I've extended `SwMgmt.select_addon_products` to receive an array of `SUSE::Connect::Remote::Service` objects. For backward compatibility reasons, it will default to the list stored in `Registration::Storage::Cache`.

PR #369, opened 2018-04-16
Support for AY autoupgrade (bsc#1087206)

lslezak:
- The old approach from SLE12 cannot add new SCC modules ([bsc#1087206](https://bugzilla.suse.com/show_bug.cgi?id=1087206))
- Reimplemented AutoYaST autoupgrade, use the same API and workflow like in a manual upgrade 
  - Adapted the _auto client to call the offline migration client (the same as used in an interactive upgrade)
  - The offline migration has been adapted to support also the AutoYaST autoupgrade mode (usually just skipping the interactive parts)
- Tested manually
- 4.0.34

## Screenshot

Tested manually with the `<confirm config:type="boolean">true</confirm>` option in the XML profile:
sles12_autoupgrade_proposal

### DIFF
--- a/src/clients/scc_auto.rb
+++ b/src/clients/scc_auto.rb
@@ -219,10 +219,8 @@ module Yast
       old = Dir[File.join(Installation.destdir, "/etc/zypp/repos.d/*")] +
         Dir[File.join(Installation.destdir, "/etc/zypp/services.d/*")]
 
-      old.each do |f|
-        log.info "Removing #{f}"
-        ::FileUtils.rm_rf(f)
-      end
+      log.info "Removing #{old}"
+      ::FileUtils.rm_rf(old)
     end
 
     # finish the registration process
--- a/src/clients/scc_auto.rb
+++ b/src/clients/scc_auto.rb
@@ -26,6 +26,7 @@
 #
 #
 
+require "fileutils"
 require "yast/suse_connect"
 
 require "registration/storage"
@@ -38,6 +39,7 @@ require "registration/connect_helpers"
 require "registration/ssl_certificate"
 require "registration/url_helpers"
 require "registration/ui/autoyast_config_workflow"
+require "registration/ui/offline_migration_workflow"
 require "registration/erb_renderer.rb"
 
 module Yast
@@ -187,9 +189,15 @@ module Yast
 
       # update the registration in AutoUpgrade mode if the old system was registered
       if Mode.update && old_system_registered?
-        updated = update_registration
-        log.info "Registration updated: #{updated}"
-        return updated
+        # drop all obsolete repositories and services (manual upgrade contains a dialog
+        # where the old repositories are deleted, in AY we need to do it automatically here)
+        # Note: the Update module creates automatically a backup which is restored
+        # when upgrade is aborted or crashes.
+        repo_cleanup
+
+        ret = ::Registration::UI::OfflineMigrationWorkflow.new.main
+        log.info "Migration result: #{ret}"
+        return ret == :next
       end
 
       ret = ::Registration::ConnectHelpers.catch_registration_errors do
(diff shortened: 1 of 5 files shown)


src/clients/scc_auto.rb
```diff
@@ -204,6 +212,19 @@ def write
...
+    # delete all previous services and repositories
+    def repo_cleanup
+      # we cannot use pkg-bindings here because loading services would trigger
+      # service and repository refresh which we want to avoid (it might easily fail)
+      old = Dir[File.join(Installation.destdir, "/etc/zypp/repos.d/*")] +
+        Dir[File.join(Installation.destdir, "/etc/zypp/services.d/*")]
+
+      old.each do |f|
+        log.info "Removing #{f}"
+        ::FileUtils.rm_rf(f)
```
imobachgs: What about calling `FileUtils.rm_rf` once? Something like: ```::FileUtils.rm_rf(old)``?

src/lib/registration/ui/migration_repos_workflow.rb
```diff
@@ -178,6 +175,23 @@ def registration_check
...
+      # @return [nil, Symbol] the workflow symbol (:next, :skip) or nil if not
+      #   in an installation
+      def registration_check_at_installation
+        # handle system upgrade (fate#323163)
+        return nil unless Yast::Stage.initial
+        # test autoupgrade first, Mode.update covers the autoupgrade as well
+        return registration_check_at_autoupgrade if Yast::Mode.autoupgrade
+        return system_upgrade_check if Yast::Mode.update
+
+        nil
```
imobachgs: np: I think `nil` is not needed here. But certainly it makes the code easier to read.

src/lib/registration/ui/migration_repos_workflow.rb
```diff
@@ -354,6 +374,18 @@ def migration_confirmed?(base_product, activations)
       # run the migration target selection dialog
       # @return [Symbol] workflow symbol (:next or :abort)
       def select_migration_products
+        if Yast::Mode.auto
```
imobachgs: I would extract this part to a separate method.

src/lib/registration/ui/migration_repos_workflow.rb
```diff
@@ -354,6 +374,18 @@ def migration_confirmed?(base_product, activations)
...
+          # TODO: for now simply select the first found migration (bsc#1087206#c7)
+          # later we can improve this (either choose the migration with higher product versions
+          # or allow selecting the migration in the AY profile)
+          self.selected_migration = migrations.first
+          self.manual_repo_selection = false
+          log.warn "More than one migration available, usign the first one" if migrations.size > 1
```
imobachgs: typo: using

test/migration_repos_workflow_spec.rb
```diff
@@ -247,6 +247,23 @@
       allow(Registration::UrlHelpers).to receive(:registration_url)
     end
 
+    context "no migration product found" do
+      before do
+        expect(Y2Packager::ProductUpgrade).to receive(:new_base_product).and_return(nil)
```
imobachgs: Putting an `expect` call looks wrong to me. You can change it to `allow` and put an `expect` in the test where want to check for that.

src/lib/registration/ui/migration_repos_workflow.rb
lslezak: Yes, I wanted to make it clear that `nil` is returned at this point.

test/migration_repos_workflow_spec.rb
lslezak: This `expect` is intended for the whole `context`, it means that "no migration product found" condition. And to make sure that context is really used I used `expect`.

imobachgs: Ok, let's keep it. But to me, checking things in the "before" block is wrong.

PR #370, opened 2018-04-27
Install a registration rollback script (bsc#1089643)

lslezak:
- The script starts automatic rollback of the registration when YaST is aborted or crashes during upgrade.
- Tested manually with DUD and the latest build, tested also the happy path (the testing system was also successfully migrated to SLE15)
- ENOTIME, no unit tests, this is urgent for RC4 :confused: 

#### Aborting Upgrade

When the upgrade is manually aborted the script restores back the original registration status.
rollback_abort

When YaST crashes hard the rollback script is also called. For the testing purposes I have exited YaST using the embedded debugger, but in reality it would be a hard crash (e.g. segfault, OOM killer,...).
rollback_crash

*(Note: the tests were run with `startshell=1`, normally the linuxrc menu would be displayed at the end.)*

### DIFF
--- a/src/data/registration/registration_rollback.sh
+++ b/src/data/registration/registration_rollback.sh
@@ -1,6 +1,8 @@
 #! /bin/sh
 
 # This script rolls back the registration status on the server
+# You can optionally pass the target chroot directory as a parameter,
+# the default is "/".
 
 ROOT=${1:-/}
 
--- a/src/data/registration/registration_rollback.sh
+++ b/src/data/registration/registration_rollback.sh
@@ -1,0 +1,12 @@
+#! /bin/sh
+
+# This script rolls back the registration status on the server
+
+ROOT=${1:-/}
+
+# print the progress in green so it is better visible
+echo -e "\e[0;32mRestoring the registration status at ${ROOT}, \
+this might take several minutes...\e[0m"
+
+# run the rollback
+chroot "$ROOT" SUSEConnect --rollback
--- a/src/lib/registration/rollback_script.rb
+++ b/src/lib/registration/rollback_script.rb
@@ -1,0 +1,72 @@
+# ------------------------------------------------------------------------------
+# Copyright (c) 2018 SUSE LLC
+#
+# This program is free software; you can redistribute it and/or modify it under
+# the terms of version 2 of the GNU General Public License as published by the
+# Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+# FOR A PARTICULAR PURPOSE.  See the GNU General Public License for more details.
+#
+# ------------------------------------------------------------------------------
+
+require "yast"
+require "fileutils"
+
+module Registration
+  # This class handles creating a registration rollback script  which is called
+  # when the upgrade is aborted or YaST crashes. The script body is stored
+  # in the data/registration/registration_rollback.sh file.
+  class RollbackScript
+    include Yast::Logger
+
+    BACKUP_DIR = "var/adm/backup/system-upgrade".freeze
+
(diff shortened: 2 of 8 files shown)


src/data/registration/registration_rollback.sh
```diff
@@ -0,0 +1,12 @@
+#! /bin/sh
+
+# This script rolls back the registration status on the server
```
jreidinger: please document that script have also one parameter

src/data/registration/registration_rollback.sh
```diff
@@ -0,0 +1,12 @@
...
+
+ROOT=${1:-/}
+
+# print the progress in green so it is better visible
+echo -e "\e[0;32mRestoring the registration status at ${ROOT}, \
+this might take several minutes...\e[0m"
+
+# run the rollback
+chroot "$ROOT" SUSEConnect --rollback
```
jreidinger: will it really work?in my testing instsys I usually have mount bind /dev. Also what happen when installation abort when /mnt is already unmounted or not yet mounted?

src/lib/registration/rollback_script.rb
```diff
@@ -0,0 +1,72 @@
...
+    def initialize(name: DEFAULT_SCRIPT_NAME, root: "/mnt")
+      @name = name
+      @root = root
+    end
+
+    # create the registration rollback script
+    # @note The script can be created only when the target root is mounted.
+    def create
+      log.info "Creating registration rollback script #{script_path}"
+      src_file = File.expand_path("../../../data/registration/registration_rollback.sh", __FILE__)
```
jreidinger: This looks wrong for me as it breaks y2dir, directory precedence and so on. It should look like
```
  Yast::Directory.find_data_file("registration/registration_rollback.sh")
```

src/lib/registration/rollback_script.rb
```diff
@@ -0,0 +1,72 @@
...
+    # they are stored in script with number 0100
+    DEFAULT_SCRIPT_NAME = "restore-0200-registration.sh".freeze
+
+    SUSE_CONNECT = "/usr/sbin/SUSEConnect".freeze
+
+    attr_reader :name, :root
+
+    # @param name [String] script name
+    # @param root [String] target root
+    def initialize(name: DEFAULT_SCRIPT_NAME, root: "/mnt")
```
jreidinger: what is usecase for diferent script name? I do not see it used, so for simplicify of API I will just remove it.

src/data/registration/registration_rollback.sh
lslezak: Added.

src/data/registration/registration_rollback.sh
lslezak: Yes, it works, see the screencasts. I just have retested it with the updated packages.

When `/mnt` is not mounted then the scripts are not found and not executed (they are stored in `/mnt/var/adm/backup/system-upgrade`). If the target is not mounted we cannot do anything about it anyway.

See https://github.com/yast/yast-installation/blob/0c251710807ee463cf4df2a0b20c875675f9ded4/startup/YaST2.call#L279-L285

src/lib/registration/rollback_script.rb
lslezak: Um, I do not think it's a problem. And we already use this solution in several places/packages.

If you set a custom `Y2DIR` then also the `lib/registration/rollback_script.rb` file is read from this custom directory. And that data file is read from the path relative to the `rollback_script.rb` script. Which means it transitively honors the `Y2DIR` value.

lslezak: Ah, `y2update` ... This is the case which I forgot...

1) this approach is used elsewhere, we would need to fix more places
2) I considered it as a standard Ruby way (but it turned out that the YaST case is more complicated)

lslezak:
OK, we can improve it later...

PR #371, opened 2018-05-08
Fixed:Offline migration against RMT sends some requests to SCC instead

schubi2:


src/lib/registration/ui/migration_repos_workflow.rb
```diff
@@ -321,12 +321,17 @@ def load_migration_products_offline(activations)
...
+        pkg_product = Yast::Pkg.ResolvableProperties(base_product.name,
+          :product, base_product.version).find_all.first
+
         remote_product = OpenStruct.new(
-          arch:         base_product.arch.to_s,
-          identifier:   base_product.name,
-          version:      base_product.version,
+          arch:            base_product.arch.to_s,
+          identifier:      base_product.name,
+          version:         base_product.version,
```
lslezak: Could we directly set the correct "version" here instead of having extra "version_version"?

If it is really needed then add a comment why we need the extra value.

test/registration_spec.rb
```diff
@@ -256,20 +266,25 @@
...
+              ], {})
 
       subject.synchronize_products([installed_sles])
     end
   end
 
   describe "#downgrade_product" do
+    before do
+      allow_any_instance_of(Registration::Registration).to receive(:connect_params)
+        .and_return(anything)
```
lslezak: Is it really possible to use `anything` as the return value? What does it actually return? To me it seems strange, I think it should be `{}` here as well...

lslezak:
And please test also with SCC to avoid regressions with the usual case...

src/lib/registration/ui/migration_repos_workflow.rb
```diff
@@ -321,12 +321,17 @@ def load_migration_products_offline(activations)
 
         return :empty unless migration_confirmed?(base_product, activations)
 
+        # evaluate pkg.product for version_version in order to use e.g. version "152 instead of "15-0"
```
lslezak: Typo "15"

src/lib/registration/sw_mgmt.rb
```diff
@@ -252,6 +254,15 @@ def self.remote_product(product, version_release: true)
       )
     end
 
+    # remove relase string from version. E.g.: "15-0" --> "15"
+    # @param product [Y2Packager::Product] product
+    # @return [String] version
+    def self.version_without_release(product)
+      pkg_product = Yast::Pkg.ResolvableProperties(product.name,
+        :product, product.version).find_all.first
```
lslezak: This `.find_all.first` looks strange, just `.first` should be enough...

PR #374, opened 2018-05-30
Adaptation for the upgrade via SMT (bsc#1094865)

lslezak:
- This PR contains adaptations needed for upgrade via SMT
- See bug https://bugzilla.suse.com/show_bug.cgi?id=1094865
- Still WIP...

### DIFF
--- a/src/lib/registration/ssl_certificate.rb
+++ b/src/lib/registration/ssl_certificate.rb
@@ -12,8 +12,8 @@ module Registration
     Yast.import "Stage"
 
     # Path to the registration certificate in the instsys
-    INSTSYS_CERT_DIR = "/etc/pki/trust/anchors/".freeze
-    INSTSYS_SERVER_CERT_FILE = "#{INSTSYS_CERT_DIR}/registration_server.pem".freeze
+    INSTSYS_CERT_DIR = "/etc/pki/trust/anchors".freeze
+    INSTSYS_SERVER_CERT_FILE = File.join(INSTSYS_CERT_DIR, "registration_server.pem").freeze
     # Path to system CA certificates
     CA_CERTS_DIR = "/var/lib/ca-certificates".freeze
 
--- a/src/lib/registration/ssl_certificate.rb
+++ b/src/lib/registration/ssl_certificate.rb
@@ -13,11 +13,25 @@ module Registration
 
     # Path to the registration certificate in the instsys
     INSTSYS_SERVER_CERT_FILE = "/etc/pki/trust/anchors/registration_server.pem".freeze
-    # the SLE11 certificate path, see
-    # https://github.com/yast/yast-registration/blob/Code-11-SP3/src/modules/Register.ycp#L296-L297
-    SLE11_SERVER_CERT_FILE = "/etc/ssl/certs/registration-server.pem".freeze
     # Path to system CA certificates
     CA_CERTS_DIR = "/var/lib/ca-certificates".freeze
+
+    # all used certificate paths, this is used during upgrade to import
+    # the old certificate into the inst-sys, put the older paths at the end
+    # so the newer paths are checked first
+    PATHS = [
+      # the YaST (SUSEConnect) current default path
+      SUSE::Connect::YaST::SERVER_CERT_FILE,
+      # RMT certificate
+      # https://github.com/SUSE/rmt/blob/b240ce577bd1637cfb57548f2741a1925cf3e4ee/public/tools/rmt-client-setup#L214
+      "/etc/pki/trust/anchors/rmt-server.pem",
+      # SMT certificate
+      # https://github.com/SUSE/smt/blob/SMT12/script/clientSetup4SMT.sh#L245
+      "/etc/pki/trust/anchors/registration-server.pem",
+      # the SLE11 path (for both YaST and the clientSetup4SMT.sh script)
+      # https://github.com/yast/yast-registration/blob/Code-11-SP3/src/modules/Register.ycp#L296-L297
+      "/etc/ssl/certs/registration-server.pem"
+    ].freeze
 
     attr_reader :x509_cert
 
--- a/src/lib/registration/ssl_certificate.rb
+++ b/src/lib/registration/ssl_certificate.rb
@@ -12,7 +12,8 @@ module Registration
     Yast.import "Stage"
 
     # Path to the registration certificate in the instsys
-    INSTSYS_SERVER_CERT_FILE = "/etc/pki/trust/anchors/registration_server.pem".freeze
+    INSTSYS_CERT_DIR = "/etc/pki/trust/anchors/".freeze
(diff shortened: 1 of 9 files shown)


src/lib/registration/sw_mgmt.rb
```diff
@@ -443,18 +444,22 @@ def self.copy_old_credentials(source_dir)
...
       end
 
-      # check for NCC credentials
-      ncc_file = File.join(source_dir, dir, "NCCcredentials")
-      copy_old_credentials_file(ncc_file)
+      Dir[File.join(source_dir, dir, "*")].each do |path|
+        # skip non-files or NCCcredentials, they are handled differently
+        next if !File.file?(path) || File.basename(path) == "NCCcredentials"
+
+        new_path = File.join(dir, File.basename(path))
```
jreidinger: what about merging those two cases? like
```
Dir[File.join(source_dir, dir, "*")].each do |path|
  # skip non-files
   next if !File.file?(path)

   new_path =  File.basename(path) == "NCCcredentials" ? SUSE::Connect::YaST::GLOBAL_CREDENTIALS_FILE : File.join(dir, File.basename(path))
```

test/sw_mgmt_spec.rb
```diff
@@ -225,30 +226,44 @@
...
+      expect(File).to receive(:exist?).with(ncc_credentials).and_return(false)
+
+      expect(subject).to receive(:`).with("cp -a " + scc_credentials + " " +
+        File.join(target_dir, "SCCcredentials"))
       expect(SUSE::Connect::YaST).to receive(:credentials).and_return(OpenStruct.new)
 
       expect { subject.copy_old_credentials(root_dir) }.to_not raise_error
     end
+
+    it "copies old SCC credentials at upgrade" do
```
jreidinger: copy and paste error, two identical states. I think rspec will complain.

test/sw_mgmt_spec.rb
```diff
@@ -225,30 +226,44 @@
...
+        .and_return([smt_credentials])
+      expect(File).to receive(:exist?).with(smt_credentials).and_return(true)
+      expect(File).to receive(:exist?).with(ncc_credentials).and_return(false)
+
+      expect(subject).to receive(:`).with("cp -a " + smt_credentials + " " +
+        File.join(target_dir, "SMT-http_smt_example_com"))
+      expect(SUSE::Connect::YaST).to receive(:credentials).and_return(OpenStruct.new)
+
+      subject.copy_old_credentials(root_dir)
+    end
```
jreidinger: and if you want to be so explicit, add also test for RMT.

lslezak: Um, that's not necessary, both RMT and SMT use dynamic name of the credentials (usually the host name). So it's enough to test anything different to the standard `SCCcredentials`.

lslezak:
Unfortunately that `private_class_method def self.foo` trick does not work, yardoc cannot parse this and fails. So Travis fails as well... :worried:

src/lib/registration/ssl_certificate.rb
```diff
@@ -175,12 +191,15 @@ def import_to_system
     # to installation system is slightly different to the one followed
     # to import certificates to a installed system.
     #
+    # @param target_path [String] where the imported certificate will be saved,
+    #   the pathe should contain the INSTSYS_CERT_DIR prefix otherwise it might
```
jreidinger: typo pathe

src/lib/registration/ssl_certificate.rb
```diff
@@ -175,12 +191,15 @@ def import_to_system
...
+    #   not work correctly.
     # @return [Boolean] true if import was successful; false otherwise.
     #
     # @see update_instsys_ca
-    def import_to_instsys
+    def import_to_instsys(target_path = self.class.default_certificate_path)
```
jreidinger: to be honest I do not like much this new param as it is confusing when seeing it in use. consider `certificate.import_to_instsys(path)` this for me looks like it imports from path and not to path. So what about named param `to`? like `def import_to_instsys(to: self.class.default_certificate_path)` so usage will looks like `certificate.import_to_instsys(to: path)` ?

src/lib/registration/sw_mgmt.rb
```diff
@@ -443,18 +444,22 @@ def self.copy_old_credentials(source_dir)
         ::FileUtils.mkdir_p(dir)
       end
 
-      # check for NCC credentials
-      ncc_file = File.join(source_dir, dir, "NCCcredentials")
-      copy_old_credentials_file(ncc_file)
+      Dir[File.join(source_dir, dir, "*")].each do |path|
```
jreidinger: note that this by default skip hidden files starting with "." is it problem here?

src/lib/registration/sw_mgmt.rb
```diff
@@ -443,18 +444,22 @@ def self.copy_old_credentials(source_dir)
...
+        end
 
-    def self.copy_old_credentials_file(file)
-      return unless File.exist?(file)
+        copy_old_credentials_file(path, new_path)
+      end
+    end
 
-      new_file = SUSE::Connect::YaST::GLOBAL_CREDENTIALS_FILE
+    private_class_method def self.copy_old_credentials_file(file, new_file)
```
jreidinger: it is used also here. Here it does not cause problem?

lslezak: No, because there is no yardoc comment... :confused:

src/lib/registration/sw_mgmt.rb
lslezak: Hidden credentials file would be very strange, I have never seen that.

src/lib/registration/ssl_certificate.rb
lslezak: Um, the usual usage looks like
```ruby
cert = SslCertificate.load_file(cert_file)
cert.import_to_instsys(target_path)
```

Where it is IMHO obvious what is loaded from and saved to.

I do not like adding a parameter name if there is only one parameter. If the name is confusing we could change the method name to `save_to_instsys(path)` or something like that.

jreidinger: well, method `save_to_instsys`  makes good sense for me.

src/lib/registration/sw_mgmt.rb
jreidinger: ok, fine. I just want to point it out.

src/lib/registration/sw_mgmt.rb
jreidinger: ok

src/lib/registration/ssl_certificate.rb
jreidinger: but just NP, so feel free to merge.

jreidinger: btw, here I do not get whole import name, when you actually exporting certificate and do not import anything.

PR #379, opened 2018-06-27
Added additional searchkeys to desktop file

schubi2:


src/desktop/add_extensions.desktop
```diff
@@ -6,6 +6,7 @@ X-SuSE-YaST-Call=scc
 X-SuSE-YaST-Argument=select_extensions
 X-SuSE-YaST-RootOnly=yes
 X-SuSE-YaST-Group=Software
+X-SuSE-YaST-Keywords=extensions,modules,software,registration,scc
```
lslezak: Maybe also `package` and `repositories` could be useful as `software` is not that much used in the Linux world IMHO...

src/desktop/customer_center.desktop
```diff
@@ -11,6 +11,7 @@ X-SuSE-YaST-AutoInstClonable=false
 X-SuSE-YaST-AutoInstRequires=lan,proxy
 X-SuSE-YaST-AutoInstSchema=registration.rnc
 X-SuSE-YaST-AutoInstResource=suse_register
+X-SuSE-YaST-Keywords=scc,software,registration
```
lslezak: NP: Maybe also `smt` and `rmt` in addition to `scc` could be added here...

PR #384, opened 2018-07-18
Sle 12 sp3 1043125

schubi2:


src/lib/registration/exceptions.rb
```diff
@@ -41,4 +41,7 @@ class SourceRestoreError < PkgError
   # generic download error
   class DownloadError < RuntimeError
   end
+
```
lslezak: Please add a description of the exception, what does it mean or when it can happen...

src/clients/scc.rb
```diff
@@ -52,6 +52,8 @@ def main
           return WFM.call("inst_scc", WFM.Args)
         rescue Registration::SourceRestoreError => e
           retry if fix_repositories(e.message)
+        rescue Registration::PkgAborted => e
+          log.info "User abort..."
```
lslezak: I'd just add a comment that the abort has been already confirmed by user and we do not need to do anything here and just exiting is fine.

PR #391, opened 2018-08-14
Filter all development versions when selecting addons

teclator:
- Trello Card: https://trello.com/c/S7syCFlB
- Bugzilla: https://bugzilla.suse.com/show_bug.cgi?id=1104450

By now I have modified only the more relevant parts using the label proposed by @fcrozat , there are still some methods [1] that could be renamed for removing almost completely any "beta" mention.

## Hide checked

hidetestingversions

# Hide unchecked

unhidetestingversions

---

[1] https://github.com/yast/yast-registration/blob/master/src/lib/registration/ui/addon_selection_base_dialog.rb#L27

### DIFF
--- a/package/yast2-registration.changes
+++ b/package/yast2-registration.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Tue Aug 14 13:10:04 UTC 2018 - knut.anderssen@suse.com
+
+- Modified addon selection filter for filtering all the development
+  versions (bsc#1104450)
+- 4.1.0
+
+-------------------------------------------------------------------
 Wed Jul 18 10:45:15 CEST 2018 - schubi@suse.de
 
 - Do not crash if getting zypp lock failed. (bnc#1043125)
--- a/package/yast2-registration.changes
+++ b/package/yast2-registration.changes
@@ -1,8 +1,8 @@
 -------------------------------------------------------------------
 Tue Aug 14 13:10:04 UTC 2018 - knut.anderssen@suse.com
 
-- Modified addon selection filter for filtering all the development
-  versions (bsc#1104450)
+- Modified addon selection filter for filtering all the testing
+  (not released) versions (bsc#1104450)
 - 4.1.0
 
 -------------------------------------------------------------------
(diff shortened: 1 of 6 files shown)


src/lib/registration/ui/addon_selection_base_dialog.rb
```diff
@@ -92,9 +92,9 @@ def content
         vbox_elements = [Left(Heading(heading))]
         available_addons = @all_addons.reject(&:registered?)
 
-        unless available_addons.empty? || available_addons.select(&:beta_release?).empty?
+        unless available_addons.empty? || available_addons.all?(&:supported_release?)
           vbox_elements.push(Left(CheckBox(Id(:filter_beta), Opt(:notify),
-            _("&Hide Beta Versions"), check_filter)))
+            _("&Hide Development Versions"), check_filter)))
```
lslezak: - Maybe  "&Hide Testing Versions" would be better, for customers they are ment for testing, development is rather for us (SUSE).
- Add `# TRANSLATORS: Checkbox label, hides alpha or beta versions (not released yet)` comment so the meaning is clear for the translators.

package/yast2-registration.changes
```diff
@@ -1,3 +1,10 @@
+-------------------------------------------------------------------
+Tue Aug 14 13:10:04 UTC 2018 - knut.anderssen@suse.com
+
+- Modified addon selection filter for filtering all the development
```
lslezak: Maybe `all testing (not released) versions` would be better, see below.

teclator:
@lslezak updated according to your suggestions

kwwii:
Proper English for the label would be "Hide Test Versions" but I think that "Hide Development Versions" sounds better.

teclator:
@kwwii thnx for checking!!, @lslezak it is ok with you?

lslezak:
Fine with me.

package/yast2-registration.changes
```diff
@@ -1,3 +1,10 @@
+-------------------------------------------------------------------
+Tue Aug 14 13:10:04 UTC 2018 - knut.anderssen@suse.com
+
+- Modified addon selection filter for filtering all the testing
```
lslezak: NP: Use "all *development* versions" also here.

teclator: My fault, I was already merging it

PR #393, opened 2018-08-21
RegistrationCode widget: use always the the custom_url for registering

teclator:
- https://bugzilla.suse.com/show_bug.cgi?id=1100199

### DIFF
--- a/src/lib/registration/widgets/registration_code.rb
+++ b/src/lib/registration/widgets/registration_code.rb
@@ -130,10 +130,7 @@ module Registration
       # run the system and the base product registration
       # @return [Boolean] true on success
       def register_system_and_base_product
-        url = UrlHelpers.registration_url
-        return false if UrlHelpers.registration_url == :cancel
-
-        registration    = Registration.new(url)
+        registration    = Registration.new(options.custom_url)
         registration_ui = RegistrationUI.new(registration)
 
         success, product_service = registration_ui.register_system_and_base_product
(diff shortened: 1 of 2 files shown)


src/lib/registration/widgets/registration_code.rb
```diff
@@ -130,10 +130,7 @@ def valid_url?
       # run the system and the base product registration
       # @return [Boolean] true on success
       def register_system_and_base_product
-        url = UrlHelpers.registration_url
-        return false if UrlHelpers.registration_url == :cancel
-
-        registration    = Registration.new(url)
+        registration    = Registration.new(options.custom_url)
```
jreidinger: so how is now handled cancel case? Will it still work?

jreidinger: I just worry if we do not break working scenario like registration on running system.

teclator: If I'm not wrong that was mainly for slp discovery but we are even not using it as we are always using the custom_url

https://github.com/yast/yast-registration/blob/916acb65309f3a0b957423d720d3e4059aecb9b7/src/lib/registration/url_helpers.rb#L79

https://github.com/yast/yast-registration/blob/916acb65309f3a0b957423d720d3e4059aecb9b7/src/lib/registration/url_helpers.rb#L116

teclator: The current problem is that we are using always the cached url once obtained, other solution could be resetting the cached url before fetching it:

https://github.com/yast/yast-registration/blob/916acb65309f3a0b957423d720d3e4059aecb9b7/src/lib/registration/url_helpers.rb#L88

PR #394, opened 2018-08-31
Better check not installed addon products (bsc#1103412)

lslezak:
- See https://bugzilla.suse.com/show_bug.cgi?id=1103412
- Some specific repositories do not provide any product
- Fixes online migration on PPC
- 3.2.16

### DIFF
--- a/test/addon_spec.rb
+++ b/test/addon_spec.rb
@@ -113,7 +113,7 @@ describe Registration::Addon do
       expect(reg_not_installed_addons.first.name).to eql(addon2.name)
     end
 
-    it "does not return not addons with not available products" do
+    it "does not return addons without available products" do
       prod = addon_generator("name" => "prod")
       registration = double(
         activated_products: [prod],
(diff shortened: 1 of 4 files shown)


test/addon_spec.rb
```diff
@@ -105,11 +105,29 @@
...
       expect(Registration::SwMgmt).to receive(:installed_products).and_return([])
+      expect(Yast::Pkg).to receive(:ResolvableProperties).with(prod2.identifier, :product, "")
+        .and_return(["status" => :available])
       reg_not_installed_addons = Registration::Addon.registered_not_installed
 
       expect(reg_not_installed_addons.size).to eql(1)
       expect(reg_not_installed_addons.first.name).to eql(addon2.name)
     end
+
+    it "does not return not addons with not available products" do
```
dgdavid: NP: there is an extra "not". Maybe _"does not return addons without available products"_

PR #398, opened 2018-09-04
Limit the regcode input field

dgdavid:
Related to https://bugzilla.suse.com/show_bug.cgi?id=1098576

### DIFF
--- a/package/yast2-registration.changes
+++ b/package/yast2-registration.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Tue Sep  4 08:34:31 UTC 2018 - dgonzalez@suse.com
+
+- Limit the registration code size to 512 characters (bsc#1098576)
+- 4.1.2
+
+-------------------------------------------------------------------
 Fri Aug 31 10:12:16 UTC 2018 - lslezak@suse.cz
 
 - Better check the not installed addon products, some specific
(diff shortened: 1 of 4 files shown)


package/yast2-registration.changes
```diff
@@ -1,3 +1,9 @@
+-------------------------------------------------------------------
+Tue Sep  4 08:34:31 UTC 2018 - dgonzalez@suse.com
+
+- Limit max length of regcode to 512 (bsc#1098576)
```
mchf: 512 of "what" ;-)

src/lib/registration/ui/base_system_registration_dialog.rb
```diff
@@ -65,6 +65,9 @@ def run
           Yast::GetInstArgs.enable_next || Yast::Mode.normal
         )
 
+        # Limit the reg_code InputField to 512 (bsc#1098576)
```
mchf: well ... a bit useless comment ;-)

mchf:
@lslezak does it make sense to also limit min length when already touching the code?

src/lib/registration/ui/base_system_registration_dialog.rb
dgdavid: Actually, it "explains" why that control was introduced, linking to the associated bug. As far as I have understood until now (reading the code), this is a quite common practice in YaST. Isn't it?

If you want, I could reword it to something like `Avoid sending too large request limiting the registration code (bsc#1098576)` or whatever other reasonable suggestion :)

lslezak: I'm fine with the current comment, the most important part is the bug number which is present.

lslezak:
:heavy_check_mark: Public Jenkins [job #9](http://ci.opensuse.org/job/yast-yast-registration-master/9/) successfully finished

PR #400, opened 2018-10-11
Forbid using CRLF in the reg. code (bsc#1111419)

lslezak:
- Forbid using CRLF in the reg. code they cannot be included in the registration code
- The code is sent in the HTTP header which does not allow using CRLF, it is used as header field separator.
- Normally you cannot enter a new line character into the input field but you can copy & paste it by mistake with the registration code
- 3.2.17

registration-crlf

### DIFF
--- a/test/base_system_registration_dialog_test.rb
+++ b/test/base_system_registration_dialog_test.rb
@@ -49,7 +49,7 @@ describe Registration::UI::BaseSystemRegistrationDialog do
           expect(Yast::UI).to receive(:QueryWidget).with(:email, :Value)
             .and_return(email)
           expect(Yast::UI).to receive(:QueryWidget).with(:reg_code, :Value)
-            .and_return(reg_code)
+            .and_return(reg_code).twice
           expect(Yast::UI).to receive(:UserInput).and_return(:next)
 
           options = Registration::Storage::InstallationOptions.instance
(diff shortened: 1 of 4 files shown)


test/base_system_registration_dialog_test.rb
```diff
@@ -90,6 +90,23 @@
         end
       end
 
+      context "when user enters an invalid regcode" do
+        # include CRLF characters which are not allowed
+        let(:reg_code) { "\nmy-reg-code\r" }
+        it "displays error popup and does not register the system" do
+          expect(Yast::UI).to receive(:QueryWidget).with(:reg_code, :Value)
```
imobachgs: np: IMHO, you are not testing this, so I would use `allow` here to make it clear. The same for the next two lines.

test/base_system_registration_dialog_test.rb
```diff
@@ -90,6 +90,23 @@
...
+            .and_return(reg_code)
+
+          expect(Yast::UI).to receive(:UserInput).and_return(:next, :abort)
+          expect(Registration::UI::AbortConfirmation).to receive(:run).and_return(true)
+
+          expect(Yast::Report).to receive(:Error).with(/Invalid registration code/)
```
imobachgs: This (and the next line) is what you really want to test. So it is fine to use `expect` here.

PR #402, opened 2018-10-12
Sle 15 ga registration popup

schubi2:


src/lib/registration/ui/base_system_registration_dialog.rb
```diff
@@ -358,17 +359,46 @@ def help_text
...
         # %{media_name} is the media name (e.g. SLE-15-Packages),
         # %{download_url} is an URL link (e.g. https://download.suse.com)
-        warning = _("Without registration, update channels will not be\n" \
-          "configured. This disables updates and security fixes.\n\n" \
-          "A full system can be installed using the\n" \
-          "%{media_name} media from %{download_url}.\n" \
-          "Without these media only a minimum system is available\n" \
-          "in this installation.") %
-          { media_name: "SLE-15-Packages", download_url: "https://download.suse.com" }
+        if media_name && !media_name.empty? &&
```
lslezak: The `nil` check is actually not needed, `ProductFeatures.GetStringFeature` returns empty string if it is not found, see [here](https://github.com/yast/yast-yast2/blob/5b9a565af09703f569cff96e7f1a96a25b4cc903/library/control/src/modules/ProductFeatures.rb#L215). Just remove the `media_name && download_url` checks and add a comment describing that so it is clear.

src/lib/registration/ui/base_system_registration_dialog.rb
```diff
@@ -358,17 +359,46 @@ def help_text
...
+            download_url && !download_url.empty?
+          warning += "\n\n" +
+            _("A full system can be installed using the\n" \
+            "%{media_name} media from %{download_url}.") %
+            { media_name: media_name, download_url: download_url }
+        elsif media_name && !media_name.empty?
```
lslezak: I'm thinking if handling of the other cases is needed for the future. It feels a bit like over engineering...

I think the additional medium will always have some name (even if a stupid one like `DVD2`), you need to be able to somehow tell the customers to use an additional medium and that's hard without a name. :grin: 

Maybe we could simply require defining both and drop these extra conditions. On the other hand this extra code is pretty small and straightforward so I'll leave the decision up to you. :smiley:

schubi2: I even do not have a testcase for that corner cases :-) --> removed.

PR #405, opened 2018-11-06
Do not report 422 errors in "silent mode"

dgdavid:
Related to [bsc#1091825](https://bugzilla.suse.com/show_bug.cgi?id=1091825)

### DIFF
--- a/package/yast2-registration.changes
+++ b/package/yast2-registration.changes
@@ -1,4 +1,12 @@
 -------------------------------------------------------------------
+Tue Nov  6 13:56:48 UTC 2018 - dgonzalez@suse.com
+
+- Do not display a connection error when automatically trying to
+  use the base product registration code for the extensions
+  (bsc#1091825).
+- 4.1.6
+
+-------------------------------------------------------------------
 Tue Oct 16 15:35:43 CEST 2018 - schubi@suse.de
 
 - Added license file to spec.
--- a/src/lib/registration/connect_helpers.rb
+++ b/src/lib/registration/connect_helpers.rb
@@ -103,9 +103,8 @@ module Registration
           check_smt_api(error_msg)
           report_error(message_prefix + _("Connection to registration server failed."), error_msg)
         when 422
-          if silent_reg_code_mismatch && e.response.body["error"] =~
-              /does not include the requested product/
-            log.info "Reg code does not work for this product, that's OK"
+          if silent_reg_code_mismatch
+            log.info "Reg code does not work for this product."
           else
             # Error popup
             report_error(message_prefix + _("Connection to registration server failed."), error_msg)
--- a/test/connect_helpers_spec.rb
+++ b/test/connect_helpers_spec.rb
@@ -113,29 +113,49 @@ describe Registration::ConnectHelpers do
       include_examples  "old registration server", JSON::ParserError.new("error message")
     end
 
-    context "error 404 is received" do
-      include_examples  "old registration server", api_error(code: 404)
-    end
+    [400, 401, 500, 42].each do |error_code|
+      exception = api_error(code: error_code)
 
-    [400, 401, 422, 500, 42].each do |error_code|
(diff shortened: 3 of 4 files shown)


src/lib/registration/connect_helpers.rb
```diff
@@ -103,9 +103,8 @@ def self.catch_registration_errors(message_prefix: "",
           check_smt_api(error_msg)
           report_error(message_prefix + _("Connection to registration server failed."), error_msg)
         when 422
-          if silent_reg_code_mismatch && e.response.body["error"] =~
-              /does not include the requested product/
-            log.info "Reg code does not work for this product, that's OK"
+          if silent_reg_code_mismatch
+            log.info "Reg code does not work for this product"
```
mvidner: We should log `e.response.body["error"]` so that we can debug things the next time the API changes under our hands.

package/yast2-registration.changes
```diff
@@ -1,3 +1,10 @@
+-------------------------------------------------------------------
+Tue Nov  6 13:56:48 UTC 2018 - dgonzalez@suse.com
+
+- Avoid an useless error message when the SCC returns a 422 error
+  but the module is in a "silent mode".
```
mvidner: Please add the bug number

test/connect_helpers_spec.rb
```diff
@@ -113,29 +113,41 @@ def api_error(code: 400, headers: {}, body: {})
...
 
-    context "'silent_reg_code_mismatch' parameter is set and a mismatch error occurs" do
-      before do
-        allow(Registration::UrlHelpers).to receive(:registration_url)
-          .and_return(SUSE::Connect::YaST::DEFAULT_URL)
+    context "error 404 is received" do
+      include_examples "old registration server", api_error(code: 404)
+    end
+
+    context "error 402 is received" do
```
mvidner: 422

test/connect_helpers_spec.rb
```diff
@@ -113,29 +113,41 @@ def api_error(code: 400, headers: {}, body: {})
...
+      context "and 'silent_reg_code_mismatch' param is not setted" do
```
mvidner: sed -e s/setted/set/ # :-D

test/connect_helpers_spec.rb
```diff
@@ -113,29 +113,41 @@ def api_error(code: 400, headers: {}, body: {})
...
+        include_examples "reports error and returns false", api_error(code: 402)
```
mvidner: 422

src/lib/registration/connect_helpers.rb
dgdavid: Is not enough with the [`log.error`](https://github.com/yast/yast-registration/blob/c39310276f37415981a883dfab64cc977a1f578c/src/lib/registration/connect_helpers.rb#L95) at the beginning?

https://github.com/yast/yast-registration/blob/c39310276f37415981a883dfab64cc977a1f578c/src/lib/registration/connect_helpers.rb#L94-L97

I will add a simple test example for it.

package/yast2-registration.changes
```diff
@@ -1,3 +1,10 @@
+-------------------------------------------------------------------
+Tue Nov  6 13:56:48 UTC 2018 - dgonzalez@suse.com
+
+- Avoid an useless error message when the SCC returns a 422 error
+  but the module is in a "silent mode" (bsc#1091825).
```
lslezak: That "silent mode" is probably a bit confusing for the end users. I'd use more user oriented text (the technical details can be found in Git sources and history).

Something like 
```
Do not display a connection error when automatically trying to use the base product
registration code for the extensions.
```

src/lib/registration/connect_helpers.rb
mvidner: Oh, you are right, I did not look that far up :)

PR #406, opened 2018-11-07
Do not report 422 errors in "silent mode"

dgdavid:
Backport of  https://github.com/yast/yast-registration/pull/405


dgdavid:
@mvidner 

Thank you! You're right, but unfortunately I did them directly in top of `master` because I thought that they were for the SP1 only. However, @kobliha remind me that they also are needed for GA.

:)

PR #407, opened 2018-11-19
improved error handling

schubi2:
The bug is over a year old. So there is no log file available anymore and the bug cannot be reproduced.
So I have added some additional info only. Better than nothing.....


src/lib/registration/connect_helpers.rb
```diff
@@ -94,28 +94,35 @@ def self.catch_registration_errors(message_prefix: "",
       rescue SUSE::Connect::ApiError => e
         log.error "Received error: #{e.response.inspect}"
         error_msg = e.message || ""
+        # TRANSLATORS: %d is an integer
+        error_code_message = _("Error code: '%d'. ") % e.code
```
lslezak: Quotation is usually used for strings, I don't think it's needed for numbers, you can easily find what was added to the message. And I'd put that on a separate line for better readability.

```suggestion
        error_code_message = _("HTTP error code: %d\n") % e.code
```

src/lib/registration/connect_helpers.rb
```diff
@@ -139,7 +146,8 @@ def self.catch_registration_errors(message_prefix: "",
         # update the message when an old SMT server is found
         check_smt_api(e.message)
 
-        report_error(message_prefix + _("Connection to registration server failed."), e.message)
+        report_error(message_prefix + _("Connection to registration server failed."),
```
lslezak: The "Connection to registration server failed" error is actually a bit misleading in this case. The connection was OK, but the data could not be parsed. So I'd replace the original message completely. (IIRC I used this message because of the text freeze at that time.)

And I'm not sure if we should mention JSON at all, this is an implementation detail and it cannot help the user to fix the problem in any way. And as we use the SUSEConnect libarry the underlying protocol theoretically might changed without YaST noticing any difference.

```ruby
report_error(message_prefix + _("Cannot parse the data from server."), e.message)
```

package/yast2-registration.changes
```diff
@@ -1,3 +1,9 @@
+-------------------------------------------------------------------
+Mon Nov 19 16:11:29 CET 2018 - schubi@suse.de
+
+- Improved error handling/messages (bsc#1060151).
```
lslezak: The error handling is the same, we just improved the messages.

PR #411, opened 2018-12-06
Use the self update version (if present) to get the available list of updates from SCC

lslezak:
Allow to use an installation media without repositories/packages for the self update, just using the `self_update_id` and the recently added `self_update_version` to query to SCC for the available list of updates.

Part of https://trello.com/c/BTk1TcxK (fate#325482)

See also related PRs:

- https://github.com/yast/yast-installation-control/pull/74
- https://github.com/yast/skelcd-control-leanos/pull/34

### DIFF
--- a/src/lib/registration/sw_mgmt.rb
+++ b/src/lib/registration/sw_mgmt.rb
@@ -120,25 +120,29 @@ module Registration
       true
     end
 
-    # Prepare a pkg-binding product hash using the first base product available
-    # as a template and with the given self_update_id as the product name.
+    # Prepare a pkg-binding product info hash
     #
-    # With a media containing multiple products it is expected that all the
-    # products use the same version and arch.
+    # If there is available products in the media, the first of them will be used as a template,
+    # keeping the self_update_id as the product name, since it is expected that all of them use the
+    # same version and arch.
     #
     # @param self_update_id [String] product name to be used for get the installer updates
-    # @return [Hash,nil] with pkg-binding format; return nil if the
-    # given self_update_id is empty or there is no base product available
-    def self.installer_update_base_product(self_update_id)
-      return if self_update_id.empty?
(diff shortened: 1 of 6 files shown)


test/sw_mgmt_spec.rb
```diff
@@ -123,36 +123,56 @@
...
 
     context "when there is no base product available" do
       let(:available_products) { [] }
 
-      it "returns nil" do
-        allow(Y2Packager::Product).to receive(:available_base_products).and_return([])
-        expect(subject.installer_update_base_product("self_update_id")).to eq(nil)
+      it "returns the product info hash" do
+        expect(result).to include(
+          "name"         => "DummyProductName",
```
lslezak: I think using `self_update_id` here is better to avoid duplication. If you we need to change it for whatever reason we will need to touch only one place.
```suggestion
          "name"         => self_update_id,
```

src/lib/registration/sw_mgmt.rb
```diff
@@ -120,25 +120,29 @@ def self.check_repositories
...
+    #                    is no base product available
+    def self.installer_update_base_product(self_update_id, self_update_version)
       base_product = Y2Packager::Product.available_base_products.first
-      return unless base_product
 
-      # filter out not needed data
+      name = self_update_id.to_s
+      version = base_product ? version_without_release(base_product) : self_update_version.to_s
+
+      return if [name, version].any?(&:empty?)
```
lslezak: This looks too complicated to me, IMHO simple

```suggestion
      return if name.empty? || version.empty?
```

looks much more readable to me. Sure, for larger lists it would make sense, but for just two values I feel the meaning is more obvious.

dgdavid:
Closing it because finally we are going ahead with a very different approach: using a fallback repository.

PR #413, opened 2018-12-14
Fixed UI display issue in the installation workflow (bsc#1117492)

lslezak:
- https://bugzilla.suse.com/show_bug.cgi?id=1117492
- Fixes a not cleared screen
- 4.1.10

## After Applying the Fix

registration

### DIFF
--- a/src/lib/registration/clients/inst_scc.rb
+++ b/src/lib/registration/clients/inst_scc.rb
@@ -322,8 +322,6 @@ module Yast
     def workflow_start
       log.debug "WFM.Args: #{WFM.Args}"
 
-      Yast::Wizard.ClearContents
-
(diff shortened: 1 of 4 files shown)


src/lib/registration/clients/inst_scc.rb
```diff
@@ -320,6 +322,8 @@ def start_workflow
     def workflow_start
       log.debug "WFM.Args: #{WFM.Args}"
 
+      Yast::Wizard.ClearContents
```
dgdavid: Just a question: Why is this needed here if now the BaseSystemRegistration dialog is ensuring to clear the content?

lslezak: Ah, right, this was my initial test, I forgot to remove it, it works without it.

PR #414, opened 2019-01-10
Properly restart the package management (bsc#1120568)

lslezak:
- Related to https://bugzilla.suse.com/show_bug.cgi?id=1120568
- Properly restart the package management to avoid using a possibly outdated configuration

The code restarts the package manager by calling (see [here](https://github.com/yast/yast-registration/blob/master/src/lib/registration/ui/migration_repos_workflow.rb#L439-L443)):

```ruby
Yast::Pkg.SourceFinishAll
Yast::Pkg.TargetFinish
SwMgmt.init
```

However, the are two problems in that `SwMgmt.init` call.

- It calls `Product.FindBaseProducts` which internally initializes the repositories. But that's wrong without initialized target, the target has been closed by the `Pkg.TargetFinish` call before. The target defines from which directory the repositories are read, we need to be sure it's set to `/mnt`. I guess it works just by accident, because of caching the original repositories, but that has been fixed in https://github.com/yast/yast-pkg-bindings/pull/111.
- The initialized repositories are not unloaded to allow correct reinitialization later. (Pkg bindings do not initialize sources if they have been already initialized so the old values might be used.)

## Fix

- Ensure the target is initialized before calling `Product.FindBaseProducts`
- Close both target and sources after that so they can be initialized later properly

### DIFF
--- a/src/lib/registration/sw_mgmt.rb
+++ b/src/lib/registration/sw_mgmt.rb
@@ -655,7 +655,7 @@ module Registration
         # will use the old value from the upgraded system which might not
         # match the new target_distro from the media and might result in ignoring
         # service repositories (bsc#1094865)
-        options = { "target_distro" => target_distribution }
+        options = { "target_distro" => target_distribution(destdir) }
         Pkg.TargetInitializeOptions(destdir, options)
       else
         Pkg.TargetInitialize(destdir)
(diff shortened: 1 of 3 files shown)


src/lib/registration/sw_mgmt.rb
```diff
@@ -655,23 +655,31 @@ def self.init_target(destdir)
...
       base_products = Product.FindBaseProducts
 
       # empty target distribution disables service compatibility check in case
       # the base product cannot be found
       target_distro = base_products ? base_products.first["register_target"] : ""
       log.info "Base product target distribution: #{target_distro.inspect}"
 
+      # close both target and sources to fully reinitialize later
+      Pkg.SourceFinishAll
+      Pkg.TargetFinish
```
jreidinger: what is performance impact? Does initialize also download and parse all repository metadata?

lslezak: There is no performance impact, libzypp has a 10 minutes timeout by default (`repo.refresh.delay = 10`). The repository metadata are not loaded (not even checked for a change on the server) if they were downloaded during the last 10 minutes.

See https://github.com/openSUSE/libzypp/blob/master/zypp.conf#L121-L137

So registering against SCC and then immediately restarting the package manager has no performance impact.

PR #415, opened 2019-01-16
Do not crash when no base product to register is found (bsc#1122011)

lslezak:
- Related to https://bugzilla.suse.com/show_bug.cgi?id=1122011
- There is very likely some other issue, the base product to register should not be `nil`
- Anyway, YaST should not crash in that situation
- 4.1.13

### DIFF
--- a/src/lib/registration/registration.rb
+++ b/src/lib/registration/registration.rb
@@ -134,6 +134,11 @@ module Registration
       # extensions for base product
       base_product = ::Registration::SwMgmt.base_product_to_register
 
+      if !base_product
+        log.warn "No base product, skipping addons"
+        return
+      end
+
       log.info "Reading available addons for product: #{base_product["name"]}"
 
       remote_product = SwMgmt.remote_product(base_product)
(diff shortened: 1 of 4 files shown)


src/lib/registration/registration.rb
```diff
@@ -130,10 +130,17 @@ def update_system(target_distro = nil)
       ret
     end
 
+    # Get the list of addons
```
dgdavid: NP: I used to leave an empty line here :)

test/registration_spec.rb
```diff
@@ -127,15 +127,19 @@
...
       remote_product = load_yaml_fixture("remote_product.yml")
-      expect(SUSE::Connect::YaST).to receive(:show_product).and_return(remote_product)
+      allow(SUSE::Connect::YaST).to receive(:show_product).and_return(remote_product)
       # no product renames defined
-      expect(Registration::SwMgmt).to receive(:update_product_renames).with({})
+      allow(Registration::SwMgmt).to receive(:update_product_renames).with({})
+    end
 
+    it "returns empty list if no base product is found" do
+      expect(Registration::SwMgmt).to receive(:base_product_to_register).and_return(nil)
```
dgdavid: NP: I like more to use `context`s, a `let` for the _base_product_ and move this "expectation" to the `before` block using an `allow` instead.

PR #421, opened 2019-02-18
Fixed "can't modify frozen String" crash (bsc#1125006)

lslezak:
- The `String` returned from gettext is frozen and cannot be modified
- Lazy initialization to make testing easier
- 3.2.17

### DIFF
--- a/test/registration/ui/not_installed_products_dialog.rb
+++ b/test/registration/ui/not_installed_products_dialog.rb
@@ -28,7 +28,7 @@ describe Registration::UI::NotInstalledProductsDialog do
       allow(subject).to receive(:_), &:freeze
     end
 
-    context "when the there is an registered but not installed product" do
+    context "when there is a registered but not installed product" do
       before do
         allow(Registration::Addon).to receive(:registered_not_installed).and_return(
           [
--- a/test/registration/ui/not_installed_products_dialog.rb
+++ b/test/registration/ui/not_installed_products_dialog.rb
@@ -1,0 +1,54 @@
+#!/usr/bin/env rspec
+# ------------------------------------------------------------------------------
+# Copyright (c) 2018 SUSE LLC, All Rights Reserved.
+#
+# This program is free software; you can redistribute it and/or modify it under
+# the terms of version 2 of the GNU General Public License as published by the
+# Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+# FOR A PARTICULAR PURPOSE. See the GNU General Public License for more details.
+# ------------------------------------------------------------------------------
+
+require_relative "../../spec_helper"
+require "registration/ui/not_installed_products_dialog"
+
+describe Registration::UI::NotInstalledProductsDialog do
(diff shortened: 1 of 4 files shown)


dgdavid:
Nice catch, @lslezak 

> The `String` returned from gettext is frozen and cannot be modified

I am still wondering how do you realized about that? I did the below unsuccessful dummy test

```ruby
require "yast"
require "yast/i18n"

class SimpleRegistrationTest
  include Yast::I18n

  def initialize
    textdomain "registration"
  end
end
```

```irb
irb(main):001:0> test = SimpleRegistrationTest.new
=> #<SimpleRegistrationTest:0x000000000138bc28 @my_textdomain=["registration"]>                      
irb(main):002:0> out = test._("<p>The addons listed below are registered but not installed: </p>")         
=> "<p>The addons listed below are registered but not installed: </p>"                                                        
irb(main):003:0> test.frozen?
=> false
```
Another question, means that that could we get more `can't modify frozen String` crashes from now on those places where the `<<` operator has been used to append text to a previous translation?

Thank you!

test/registration/ui/not_installed_products_dialog.rb
```diff
@@ -0,0 +1,54 @@
...
+      allow(Yast::UI).to receive(:SetFocus)
+
+      allow(Yast::Popup).to receive(:Feedback).and_yield
+      allow(subject).to receive(:handle_dialog)
+
+      # the translated strings are frozen
+      allow(subject).to receive(:_), &:freeze
+    end
+
+    context "when the there is an registered but not installed product" do
```
dgdavid: NP: "When ~~the~~ there is a~~n~~ registered but not installed product"

lslezak: Ah, I edited the text, but forgot some leftovers... thanks!

lslezak:
> I am still wondering how do you realized about that? I did the below unsuccessful dummy test

It depends whether the translation for the message is available or not.

### Translation Available

```console
> rpm -q yast2-trans-cs
yast2-trans-cs-84.87.20180514.157a0650d-lp150.1.1.noarch
> LC_ALL=cs_CZ.UTF-8 ruby -ryast -e 'include Yast::I18n; textdomain "base"; puts _("&Yes"); puts _("&Yes").frozen?'
&Ano
true
```

So the result is a frozen translated string.

### Translation Not Available

```console
> LC_ALL=C rpm -q yast2-trans-es
package yast2-trans-es is not installed
> LC_ALL=es_ES.UTF-8 ruby -ryast -e 'include Yast::I18n; textdomain "base"; puts _("&Yes"); puts _("&Yes").frozen?'
&Yes
false
```

In this case you'll get the original untranslated and *not frozen* string.

The problem is that when running the unit tests in Travis, Jenkins or OBS no translation is installed and the tests run in English locale. However, on users system the translations are usually available. That means it's hard to test it.

I'll open the topic at yast-devel@, a possible solution would be to change

```diff
- found ? Translation._(str) : str
+ found ? Translation._(str) : str.freeze
```
at https://github.com/yast/yast-ruby-bindings/blob/4652448963160abc234d1199f06bf10f08989cf5/src/ruby/yast/i18n.rb#L63 so it behaves consistently.

PR #428, opened 2019-03-10
Added option to retry the block given

teclator:
## Problem

When an exception or error occurs in the qa SCC proxy, the shown description is not user friendly, it shows all the text in the dialog and only permits to accept it.

- https://trello.com/c/G5X7kbBy/750-3-5-sles15-p2-1058375-build-2311-openqa-test-fails-in-sccregistration-connection-time-out-on-scc-registration
- https://bugzilla.suse.com/show_bug.cgi?id=1058375 
- https://bugzilla.suse.com/show_bug.cgi?id=1126045

## Solution

- Added an option to connection_helper for retrying the given blockn when some exception occurs.
- Added a details button to show the details of the occurred error in the retry dialog.

## Testing

- *Added uni test*
- *Tested manually*

## Screenshots

Screenshot_testing_2019-03-13_09:06:14

Screenshot_testing_2019-03-11_00:08:39

### DIFF
--- a/src/lib/registration/connect_helpers.rb
+++ b/src/lib/registration/connect_helpers.rb
@@ -63,6 +63,7 @@ module Registration
     def self.catch_registration_errors(message_prefix: "",
       show_update_hint: false,
       silent_reg_code_mismatch: false,
+      retry_block: false,
       &block)
       # import the SSL certificate just once to avoid an infinite loop
       certificate_imported = false
@@ -82,11 +83,16 @@ module Registration
         log.error "Timeout error: #{e.message}"
         # FIXME: to not break existing translation, this typo should be fixed
         # later after SP2: time -> timed
-        Yast::Report.Error(
-          error_with_details(message_prefix + _("Connection time out.") + "\n",
+        if retry_block
+          retry if report_error_and_retry?(message_prefix + _("Connection time out."),
             _("Make sure that the registration server is reachable and\n" \
               "the connection is reliable."))
-        )
+        else
+          report_error(message_prefix + _("Connection time out.") + "\n",
+            _("Make sure that the registration server is reachable and\n" \
+              "the connection is reliable."))
+        end
+
         false
       rescue SUSE::Connect::ApiError => e
         log.error "Received error: #{e.response.inspect}"
@@ -142,7 +148,12 @@ module Registration
         log.error "JSON parse error"
         # update the message when an old SMT server is found
         check_smt_api(e.message)
-        report_error(message_prefix + _("Cannot parse the data from server."), e.message)
+        msg = message_prefix + _("Cannot parse the data from server.")
+        if retry_block
+          retry if report_error_and_retry?(msg, e.message)
+        else
+          report_error(msg, e.message)
+        end
       rescue StandardError => e
         log.error("SCC registration failed: #{e.class}: #{e}, #{e.backtrace}")
         Yast::Report.Error(
@@ -155,6 +166,15 @@ module Registration
 
     def self.report_error(msg, error_message)
       Yast::Report.Error(error_with_details(msg, error_message))
+    end
+
+    def self.retry_error(msg, error_message)
+      buttons = { retry: _("Retry"), cancel: _("Cancel") }
+      Yast2::Popup.show(msg, details: error_message, headline: :error, buttons: buttons)
+    end
+
+    def self.report_error_and_retry?(msg, details_message)
+      retry_error(msg, details_message) == :retry
     end
 
     # Report a pkg-bindings error. Display a message with error details from
@@ -338,6 +358,7 @@ module Registration
     end
 
     private_class_method :report_error, :error_with_details, :import_ssl_certificate,
-      :report_ssl_error, :check_smt_api, :handle_network_error
+      :report_ssl_error, :check_smt_api, :handle_network_error, :retry_error,
+      :report_error_and_retry?
   end
 end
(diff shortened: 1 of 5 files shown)


src/lib/registration/connect_helpers.rb
```diff
@@ -157,6 +168,15 @@ def self.report_error(msg, error_message)
       Yast::Report.Error(error_with_details(msg, error_message))
     end
 
+    def self.retry_error(msg, error_message)
+      buttons = { retry: _("Retry"), cancel: _("Cancel") }
+      Yast2::Popup.show(msg, details: error_message, headline: :error, buttons: buttons)
```
lslezak: We really need to use `Yast::Report` module here, the `Yast2::Popup` module does not support AutoYaST mode.

Or we should handle it here, e.g. in AY mode use `Yast::Report` (without retrying, if that's not possible) and in the interactive mode use `Yast2::Popup`.

src/lib/registration/connect_helpers.rb
```diff
@@ -63,6 +63,7 @@ class ConnectHelpers
     def self.catch_registration_errors(message_prefix: "",
       show_update_hint: false,
       silent_reg_code_mismatch: false,
+      retry_block: false,
```
lslezak: Do we need an extra parameter to enable it? I mean, in my opinion it makes sense always retrying...

src/lib/registration/connect_helpers.rb
```diff
@@ -142,7 +148,12 @@ def self.catch_registration_errors(message_prefix: "",
         log.error "JSON parse error"
         # update the message when an old SMT server is found
         check_smt_api(e.message)
-        report_error(message_prefix + _("Cannot parse the data from server."), e.message)
+        msg = message_prefix + _("Cannot parse the data from server.")
```
lslezak: I'm not sure if we should retry on `JSON::ParserError`, the server might register the system properly (i.e. eating one of the customer subscription, but only the response is malformed or corrupted when transmitting over network).

But as this was mainly an issue with the SCCProxy then let the user decide, this should not happen with normal SCC...

src/lib/registration/connect_helpers.rb
teclator: I was not sure if was fine in all the scenarios, so added it as an option (probably if silent_reg_code_mismatch is true then the retry should be skipped).

If it would be nice to have it always, then yes, the parameter is not needed.

src/lib/registration/connect_helpers.rb
teclator: > I'm not sure if we should retry on `JSON::ParserError`, the server might register the system properly (i.e. eating one of the customer subscription, but only the response is malformed or corrupted when transmitting over network).

MMM in that case, when retrying the registration, it will be skipped as detected as already register, isn't it?

> 
> But as this was mainly an issue with the SCCProxy then let the user decide, this should not happen with normal SCC...

Yes, this looks more like the old registration server error than a normal SCC one, although the notes in Trello suggested to add the [Retry] option

lslezak: I'm fine with adding retry here, my point is that the user should know the consequences... But that's actually the case also for the timeout exception. It might happen that the server registers the client correctly but because of some network issue the client gets timeout...

src/lib/registration/connect_helpers.rb
lslezak: `silent_reg_code_mismatch` is for automatic trying the base product reg. code also for addons (if fails we ask the user for the reg code). But if it times out here then I think we should report it to the user as it is very likely that it would timeout also with the user provided reg. code.

lslezak: To make it short: retry always, user can always press "cancel" if it is not desired.

src/lib/registration/connect_helpers.rb
jreidinger: Ideal solution for me is to have `Yast2::Report` but for that I need a bit time. Basic idea is to have same syntax as Yast2::Popup, just add param with severity and also setup Report configuration for that new Report. So not hard, but need time to implement.

src/lib/registration/connect_helpers.rb
teclator: Exactly, and the timeout could happen at any time, that was what happened with the SCC qa proxy but there also the error is handled incorrectly.

teclator: I removed the retry, but could add it back if wanted.

src/lib/registration/connect_helpers.rb
```diff
@@ -142,6 +142,7 @@ def self.catch_registration_errors(message_prefix: "",
         # update the message when an old SMT server is found
         check_smt_api(e.message)
         details_error(message_prefix + _("Cannot parse the data from server."), e.message)
```
teclator: while Yast::Report returned `nil`, details_error reports `:ok` so, added the return value explicitly as we have in the other rescued exceptions

src/lib/registration/registration_ui.rb
```diff
@@ -55,7 +55,7 @@ def initialize(registration)
     def register_system_and_base_product
       product_service = nil
       # TRANSLATORS: Popup error message prefix
-      error_options = { message_prefix: _("Registration failed.") }
+      error_options = { message_prefix: _("Registration failed.") + "\n\n" }
```
teclator: Added new lines to the message prefix:

Screenshot_testing_2019-03-13_09:06:14

PR #433, opened 2019-05-29
Replace SMT occurrences by RMT or SMT/RMT

dgdavid:
## Problem

* https://bugzilla.suse.com/show_bug.cgi?id=1136433
* https://trello.com/c/O1hyJX01/
* Related to #431

Although from SLE-15 onward [only RMT is offered](https://github.com/SUSE/rmt), there are some parts that still referring to [SMT](https://github.com/SUSE/smt). Namely, 

* in [widget labels](https://github.com/yast/yast-registration/blob/0944ebf458ee04517bc3441ad96357a3c8115c62/src/lib/registration/widgets/registration_code.rb#L38)
* in [help texts](https://github.com/yast/yast-registration/blob/0944ebf458ee04517bc3441ad96357a3c8115c62/src/lib/registration/widgets/registration_code.rb#L38)
* in comments

## Solution

Replace `SMT` by `RMT`. Or simply add `RMT` (mostly in comments).

## Other considerations

Read the comments below, regarding to `Yast::Registration::SmtStatus` class.

### DIFF
--- a/test/sw_mgmt_spec.rb
+++ b/test/sw_mgmt_spec.rb
@@ -294,15 +294,15 @@ describe Registration::SwMgmt do
       subject.copy_old_credentials(root_dir)
     end
 
-    it "copies old SMT credentials at upgrade" do
-      smt_credentials = File.join(root_dir, target_dir, "SMT-http_smt_example_com")
+    it "copies old SMT/RMT credentials at upgrade" do
+      credentials = File.join(root_dir, target_dir, "RMT-http_example_com")
       allow(Dir).to receive(:[]).with(File.join(root_dir, target_dir, "*"))
-        .and_return([smt_credentials])
+        .and_return([credentials])
 
       allow(SUSE::Connect::YaST).to receive(:credentials).and_return(OpenStruct.new)
 
       expect(Yast::Execute).to receive(:locally!)
-        .with("cp", "-a", smt_credentials, File.join(target_dir, "SMT-http_smt_example_com"))
+        .with("cp", "-a", credentials, File.join(target_dir, "RMT-http_example_com"))
 
       subject.copy_old_credentials(root_dir)
     end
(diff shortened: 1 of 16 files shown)


dgdavid:
In the Trello card is stated, 

> Also, we have not stated whether we want to update
> 
> - class names? Registration::SmtStatus

However, in spite of SLE-15 seems to only support RMT (see https://github.com/SUSE/rmt#rmt-and-smt), I am not sure if the [`src/lib/registration/smt_status.rb`](https://github.com/yast/yast-registration/blob/master/src/lib/registration/smt_status.rb) should be moved to `src/lib/registration/rmt_status.rb` since it is checking a NCC api which it is not available for RMT [see https://github.com/SUSE/rmt/issues/103]

Thoughts?

dgdavid:
> ...
> I am not sure if the `src/lib/registration/smt_status.rb` should be moved to `src/lib/registration/rmt_status.rb` ...
>

Discussed with @lslezak in IRC

> \<lslezak> dgdavid: I'd keep it as it is, it checks for the old SMT with the old NCC API
> \<lslezak> dgdavid: just in case someone uses the old server by mistake
> \<lslezak> dgdavid: later (SLE16?) we could drop it completely

test/sw_mgmt_spec.rb
```diff
@@ -294,15 +294,15 @@
       subject.copy_old_credentials(root_dir)
     end
 
-    it "copies old SMT credentials at upgrade" do
-      smt_credentials = File.join(root_dir, target_dir, "SMT-http_smt_example_com")
+    it "copies old SMT/RMT credentials at upgrade" do
+      credentials = File.join(root_dir, target_dir, "SMT-RMT-http_example_com")
```
lslezak: NP: probably it should be just `RMT-http_example_com`, but it actually does not matter as YaST copied everything regardless the name.

PR #438, opened 2019-06-12
Improve error message for offline upgrade

dgdavid:
## Problem

During the offline upgrade, the installer **seems to** get stuck in a loop during when there is some product not activated yet.

https://bugzilla.suse.com/show_bug.cgi?id=1078739

## Actually

>  it's not the installer which is looping **but the user insisting on doing an impossible upgrade**
>
> [@lslezak in the 5th comment](https://bugzilla.suse.com/show_bug.cgi?id=1078739#c5)

## Solution

Improve the error message, changing it from simply

> No migration product found.

to 

> No migration product found.
>
> Please, boot the original system and make sure that it and all registerable products are correctly registered.
> Also check that the installed system is supported for upgrade to this _product display_name_.

## Screenshot

---

See also https://github.com/yast/yast-registration/pull/437 for a better understanding why there is not a clear and better solution for this

> [quite artificial situation](https://github.com/yast/yast-registration/pull/437)

### DIFF
--- a/src/lib/registration/ui/migration_repos_workflow.rb
+++ b/src/lib/registration/ui/migration_repos_workflow.rb
@@ -335,11 +335,21 @@ module Registration
       def load_migrations_for_products(products, remote_product)
         log.info "Loading offline migrations for target product: #{remote_product.inspect}"
         log.info "Installed products: #{products.inspect}"
+
         self.migrations = registration_ui.offline_migration_products(products, remote_product)
 
         if migrations.empty?
-          # TRANSLATORS: Error message
-          Yast::Report.Error(_("No migration product found."))
+          msg = [
+            # TRANSLATORS. Error message
+            _("No migration product found."),
+            # TRANSLATORS: Help message
+            _("Please, boot the original system and make sure " \
+              "that it and all registerable products are correctly registered.\n" \
+              "Also check that the installed system is supported for upgrade to this new product.")
+          ]
+
+          Yast::Report.Error(msg.join("\n\n"))
+
           return Yast::Mode.auto ? :abort : :empty
         end
 
--- a/src/lib/registration/ui/migration_repos_workflow.rb
+++ b/src/lib/registration/ui/migration_repos_workflow.rb
@@ -342,10 +342,11 @@ module Registration
           msg = [
             # TRANSLATORS. Error message
             _("No migration product found."),
-            # TRANSLATORS: Help message
+            # TRANSLATORS: Help message, %{product} is the product name
             _("Please, boot the original system and make sure " \
-              "that it and all registerable products are correctly registered.\n" \
-              "Also check that the installed system is supported for upgrade to this new product.")
+              "that all registerable products are correctly registered.\n" \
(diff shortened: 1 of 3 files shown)


src/lib/registration/ui/migration_repos_workflow.rb
```diff
@@ -335,11 +335,21 @@ def load_migration_products_offline(activations)
...
 
         if migrations.empty?
-          # TRANSLATORS: Error message
-          Yast::Report.Error(_("No migration product found."))
+          msg = [
+            # TRANSLATORS. Error message
+            _("No migration product found."),
+            # TRANSLATORS: Help message
+            _("Please, boot the original system and make sure " \
+              "that it and all registerable products are correctly registered.\n" \
```
lslezak:
```suggestion
              "that all registerable products are correctly registered.\n" \
```

src/lib/registration/ui/migration_repos_workflow.rb
```diff
@@ -335,11 +335,21 @@ def load_migration_products_offline(activations)
...
+              "Also check that the installed system is supported for upgrade to this new product.")
```
lslezak: Maybe we could use the real product name and replace the generic "this new product" words.

```suggestion
              "Also check that the installed system is supported for upgrade to \n%{product}.") % {product: remote_product}
```

src/lib/registration/ui/migration_repos_workflow.rb
```diff
@@ -335,11 +335,22 @@ def load_migration_products_offline(activations)
...
-          # TRANSLATORS: Error message
-          Yast::Report.Error(_("No migration product found."))
+          msg = [
+            # TRANSLATORS. Error message
+            _("No migration product found."),
+            # TRANSLATORS: Help message, %{product} is the product name
+            _("Please, boot the original system and make sure " \
+              "that all registerable products are correctly registered.\n" \
+              "Also check that the installed system is supported for upgrade to \n" \
+              "%{product}.") % { product: remote_product }
```
dgdavid: @lslezak 

`remote_product` is an `OpenStruct`. The message looks _weird_.

Screenshot_openSUSE-Leap_2019-06-13_09:06:06


What do you think about change it to something more human readable?

Like

```ruby
{ product: "#{remote_product.identifier} #{remote_product.version}"
```

which will produce the `SLED 15.1` output.

lslezak: Oh, I overlooked that...

Then I'd suggest using the `display_name` attribute, it contains a human readable product name like "SUSE Linux Enterprise Server 15 SP1" which is better than the internal identifier.

```suggestion
              "%{product}.") % { product: Y2Packager::ProductUpgrade.new_base_product.display_name }
```

(Not tested, please test it to make sure it really works as expected. :wink: )

dgdavid: Nice! It works.

I'll update the `Screenshot` section in the PR's description.

Thanks!

PR #439, opened 2019-06-18
Do not abort the installation when an addon EULA is refused

dgdavid:
## Problem

When user refuses an addon's EULA (doing click in "Next" without check the _I Agree to the License Terms._) the installation is completely aborted.

* https://bugzilla.suse.com/show_bug.cgi?id=1114018
* https://trello.com/c/Kc9mRVpB

## Solution

In spite of in the current code there is an intentional (and broken) behavior to

>  go back or abort if any EULA has not been accepted — [addon_eula_dialog.rb#L63](https://github.com/yast/yast-registration/blob/e25b04d710ff0d2a688dcd6d5a050fbc361b4580/src/lib/registration/ui/addon_eula_dialog.rb#L63)

it seems that the correct behavior should be **just go ahead _refusing_ the license and not installing/registering the addon**, according to the message shown to the user.

<details>
<summary>Show/hide screenshot</summary>

---

Screenshot_sles12-sp4_2019-06-19_11_28_47

</details>

To do that, it has been necessary to update `yast2-packager` too (see https://github.com/yast/yast-packager/pull/452), in order to support "refuse" as a cancel action. Otherwise, it is not possible to distinguish when the "abort" means "to refuse the license agreement" or "abort the installation".

## Tests

* Added and updated some unit tests.
* Tested manually via driver update.

## Caveats

The selection is untouched. I.e., when a license is refused but the user goes back to the select extensions dialog, the product is still selected.

### DIFF
--- a/src/clients/inst_scc.rb
+++ b/src/clients/inst_scc.rb
@@ -204,7 +204,7 @@ module Yast
 
     # display EULAs for the selected addons
     def addon_eula
-      ::Registration::UI::AddonEulaDialog.run(@selected_addons)
+      ::Registration::UI::AddonEulaDialog.run(Registration::Addon.selected)
     end
 
     # remember the user entered values so they can be stored to the AutoYast
--- a/src/clients/inst_scc.rb
+++ b/src/clients/inst_scc.rb
@@ -135,9 +135,6 @@ module Yast
       # FIXME: available_addons is called just to fill cache with popup
       return :cancel if get_available_addons == :cancel
 
-      # FIXME: workaround to reference between old way and new storage in Addon metaclass
-      @selected_addons = Registration::Addon.selected
-      ::Registration::Storage::InstallationOptions.instance.selected_addons = @selected_addons
       Registration::UI::AddonSelectionRegistrationDialog.run(@registration)
     end
 
--- a/src/clients/inst_scc.rb
+++ b/src/clients/inst_scc.rb
@@ -161,7 +161,8 @@ module Yast
     # back returns directly to the extensions selection
     def register_addons
       return false if init_registration == :cancel
-      ret = registration_ui.register_addons(@selected_addons, @known_reg_codes)
+
+      ret = registration_ui.register_addons(Registration::Addon.to_register, @known_reg_codes)
       ret = :extensions if ret == :back
       ret
     end
--- a/src/lib/registration/ui/addon_eula_dialog.rb
+++ b/src/lib/registration/ui/addon_eula_dialog.rb
@@ -129,13 +129,18 @@ module Registration
 
           eula_reader = EulaReader.new(tmpdir)
           setup_eula_dialog(addon, eula_reader, tmpdir)
+          result = run_eula_dialog(eula_reader)
 
-          response = run_eula_dialog(eula_reader)
-
-          addon.accept_eula if response == :accepted
-          addon.refuse_eula if response == :refuse
-
-          [:accepted, :refuse].include?(response) ? :next : response
+          case result
+          when :accepted
+            addon.accept_eula
+            :next
+          when :refuse
+            addon.refuse_eula
+            :next
+          else
+            result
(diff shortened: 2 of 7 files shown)


src/clients/inst_scc.rb
```diff
@@ -136,8 +136,8 @@ def select_addons
       return :cancel if get_available_addons == :cancel
 
       # FIXME: workaround to reference between old way and new storage in Addon metaclass
-      @selected_addons = Registration::Addon.selected
-      ::Registration::Storage::InstallationOptions.instance.selected_addons = @selected_addons
+      # @selected_addons = Registration::Addon.selected
+      # ::Registration::Storage::InstallationOptions.instance.selected_addons = @selected_addons
```
lslezak: No dead code please, remove it completely. If needed it can be restored from the Git history.

src/lib/registration/ui/addon_eula_dialog.rb
```diff
@@ -116,23 +112,35 @@ def setup_eula_dialog(addon, eula_reader, tmpdir)
...
+          eula_reader = EulaReader.new(tmpdir)
           setup_eula_dialog(addon, eula_reader, tmpdir)
-          run_eula_dialog(eula_reader)
+          result = run_eula_dialog(eula_reader)
+
+          case result
+          when :accepted
+            addon.accept_eula
+            :next
+          when :refuse
```
lslezak: Um, the results should be symmetric, if "accept**ed**" already exists it should be "refuse**d**".

test/addon_spec.rb
```diff
@@ -130,6 +225,86 @@
...
+      context "and the license was accepted" do
+        it "returns false" do
+          addon.accept_eula
+
+          expect(addon.eula_refused?).to eq(false)
+        end
+      end
+
+      context "and the license was refused" do
+        it "returns true" do
```
lslezak: Ehm, does not seem so...

test/addon_spec.rb
```diff
@@ -130,6 +225,86 @@
...
+  end
+
+  context "#eula_acceptance_needed" do
+    let(:eula_url) { nil }
+
+    before do
+      allow(addon).to receive(:eula_url).and_return(eula_url)
+    end
+
+    context "when there is not an EULA url" do
```
lslezak: Just a better wording

```suggestion
    context "when the EULA URL is nil" do
```

test/registration/ui/addon_eula_dialog_spec.rb
```diff
@@ -0,0 +1,133 @@
...
+    allow(eula_dialog).to receive(:download_eula).and_return(true)
+    registered_addon.registered
+  end
+
+  describe "#run" do
+    before do
+      allow(Yast::Wizard).to receive(:SetContents)
+    end
+
+    context "when there are not EULA acceptances to show" do
```
lslezak:
```suggestion
    context "when there are no EULA acceptances to show" do
```

src/clients/inst_scc.rb
dgdavid: Ooops! This is a clear leftover. I thought that removed it before make the commit... but... 

Thanks!

src/lib/registration/ui/addon_eula_dialog.rb
dgdavid: You're right! I'll change it.

test/addon_spec.rb
dgdavid: :man_facepalming: You know, copy&paste.

Anyway, I have new changes that will remove this part.

lslezak: No problem, that's why we have the code review. :smiley:

dgdavid:
`yast2-packager` dependency updated. After [resolve a conflict the new version is `3.3.1`](https://github.com/yast/yast-packager/pull/452#issuecomment-507179107).

PR #440, opened 2019-07-01
Keep the addons EULA workflow synced 

dgdavid:
This is a request for comments just to know if we must/should submit changes "synced" with those introduced in #439 for SLE-12-SP5.

The thing is that the _same_ fix for SLE-15 does not need these changes. Instead, the `yast2-addon` module has been updated (see yast/yast-add-on#78). So, it could be said/it seems to me that this is a dead code in SLE-15 branches, but I'm not 100% sure.

That's why I'm requesting to, at least, keep it "updated". Although any suggestion or clarification is, of course, **very** welcome.

### DIFF
--- a/test/registration/ui/addon_eula_dialog_test.rb
+++ b/test/registration/ui/addon_eula_dialog_test.rb
@@ -15,13 +15,129 @@ require_relative "../../spec_helper"
 require "registration/ui/addon_eula_dialog"
 
 describe Registration::UI::AddonEulaDialog do
-  subject(:dialog) { described_class.new([addon]) }
+  subject(:dialog) { described_class.new(addons) }
 
-  let(:addon) do
-    addon_generator("name" => "SLES", "eula_url" => "https://suse.com/download/SLES/eula")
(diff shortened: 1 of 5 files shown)


test/registration/ui/addon_eula_dialog_test.rb
```diff
@@ -15,13 +15,129 @@
...
+      allow(Yast::Wizard).to receive(:SetContents)
+      allow(dialog).to receive(:find_license).and_return(product_license)
+      allow(dialog).to receive(:download_eula).and_return(true)
+      registered_addon.registered
+    end
+
+    context "when there are no EULA acceptances to show" do
+      let(:addons) { [registered_addon, addon_wo_eula] }
+
+      it "does not display the eula dialog" do
```
lslezak: NP: the casing should be unified
```suggestion
      it "does not display the EULA dialog" do
```

dgdavid:
I updated the description of commit https://github.com/yast/yast-registration/pull/440/commits/1c7d041f6b3e71dce6a6f6062f10b1d6f0121754 a little bit, just to clarify why I'm commiting this changes.

PR #447, opened 2019-10-08
Fixed medium type detection (bsc#1152694)

lslezak:
- Detect the installation medium type only during installation
- It crashes in an installed system
- https://bugzilla.suse.com/show_bug.cgi?id=1152694
- Related to PR #444 
- 4.2.11

### DIFF
--- a/src/lib/registration/sw_mgmt.rb
+++ b/src/lib/registration/sw_mgmt.rb
@@ -170,7 +170,7 @@ module Registration
       # just for debugging:
       return FAKE_BASE_PRODUCT if ENV["FAKE_BASE_PRODUCT"]
 
-      return online_base_product if Y2Packager::MediumType.online?
+      return online_base_product if Stage.initial && Y2Packager::MediumType.online?
 
       # use the selected product if a product has been already selected
       selected = product_selected? if Stage.initial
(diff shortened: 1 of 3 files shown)


src/lib/registration/sw_mgmt.rb
```diff
@@ -170,7 +170,7 @@ def self.find_base_product
       # just for debugging:
       return FAKE_BASE_PRODUCT if ENV["FAKE_BASE_PRODUCT"]
 
-      return online_base_product if Y2Packager::MediumType.online?
+      return online_base_product if Stage.initial && Y2Packager::MediumType.online?
```
jreidinger: are you sure that here it is proper place? I expect fix to be in MediumType so it does not crash outside insts-sys.

lslezak: Actually the [exception](https://github.com/yast/yast-packager/blob/master/src/lib/y2packager/medium_type.rb#L86) is thrown there deliberately. Sure, the code in `Y2Packager::MediumType.online?` could return `false` (or `nil`) when not running in inst-sys. But that could hide the problem and the code could behave unexpectedly in some cases. It would be also difficult to debug. So I decided to make it fail verbosely.

PR #448, opened 2019-10-10
Fix autoyast config crash

jreidinger:


test/registration/ui/autoyast_addon_dialog_rest.rb
```diff
@@ -0,0 +1,41 @@
...
+  subject(:dialog) { described_class.new(addons) }
+
+  let(:addons) { [addon] }
+
+  let(:addon) do
+    {
+      "name"         => "free_cool_in_addon",
+      "version"      => "666",
+      "arch"         => "s390x",
+      "release_type" => "the best",
```
lslezak: The current products usually do not define any and it's `nil` so I'd use this value to make sure it does not crash somewhere.

And I'd use some nicer version number, like `42` :wink:

PR #452, opened 2019-10-18
AutoYaST support for the OnlineOnly installation medium (jsc#SLE-7214)

lslezak:
- https://trello.com/c/nkA4Os8Y/1377-5-autoyast-installation-using-the-online-media
- https://jira.suse.com/browse/SLE-7214
- Tested manually
- 4.2.13

### DIFF
--- a/src/clients/scc_auto.rb
+++ b/src/clients/scc_auto.rb
@@ -41,6 +41,8 @@ require "registration/url_helpers"
 require "registration/ui/autoyast_config_workflow"
 require "registration/ui/offline_migration_workflow"
 require "registration/erb_renderer.rb"
+require "y2packager/product_control_product"
+require "y2packager/medium_type"
 
 module Yast
   class SccAutoClient < Client
@@ -76,6 +78,9 @@ module Yast
       Yast.import "Label"
       Yast.import "Report"
       Yast.import "Popup"
+      Yast.import "Profile"
+      Yast.import "Packages"
+      Yast.import "Report"
       Yast.import "Installation"
     end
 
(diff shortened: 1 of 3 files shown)


src/clients/scc_auto.rb
```diff
@@ -212,6 +223,35 @@ def write
...
+      # import the GPG keys before refreshing the repositories
+      Yast::Packages.ImportGPGKeys
+
+      products = Y2Packager::ProductControlProduct.products
+      ay_product = Profile.current.fetch("software", {}).fetch("products", []).first
+
+      if !ay_product
+        # TRANSLATORS: error message, %s is the XML path, e.g. "software/products"
+        Report.Error(_("Missing product specification in the %s section") % "software/products")
+        return false
```
jreidinger: I prefer if specially designed class for autoyast issues will be used. I think this one with fatal severity is suitable https://github.com/yast/yast-autoinstallation/blob/master/src/lib/autoinstall/autoinst_issues/missing_value.rb

src/clients/scc_auto.rb
```diff
@@ -212,6 +223,35 @@ def write
...
+      end
+
+      control_product = products.find { |p| p.name == ay_product }
+
+      if !control_product
+        # TRANSLATORS: error message, %s is a product ID, e.g. "SLES"
+        Report.Error(_("Product %s not found") % ay_product)
```
jreidinger: and here issue with invalid value.

src/clients/scc_auto.rb
lslezak: Unfortunately the AY side expects only booleans [here](https://github.com/yast/yast-autoinstallation/blob/master/src/lib/autoinstall/autosetup_helpers.rb#L88-L95). Supporting custom AY issue would require bigger changes.

I'd keep it simple for now.

src/clients/scc_auto.rb
lslezak: :question: :question: :question: Um, what exactly?

jreidinger: I just mean different type of autoyast issue https://github.com/yast/yast-autoinstallation/blob/master/src/lib/autoinstall/autoinst_issues/invalid_value.rb

but if it is not yet so ready, feel free to keep it as it is.

PR #454, opened 2019-10-24
Online upgrade

jreidinger:
trello: https://trello.com/c/q3vLqEhT/1366-5-system-upgrade-using-the-online-media

for now can do migration of product. Failing later.

### DIFF
--- a/src/lib/registration/registration.rb
+++ b/src/lib/registration/registration.rb
@@ -138,10 +138,10 @@ module Registration
     # @return [Array<Addon>] List of addons, empty if no base product is found
     def get_addon_list(product = :to_register)
       # extensions for base product
-      base_product = if product == :to_register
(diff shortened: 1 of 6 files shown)


src/lib/registration/registration.rb
```diff
@@ -136,7 +136,11 @@ def update_system(target_distro = nil)
     # @return [Array<Addon>] List of addons, empty if no base product is found
     def get_addon_list
       # extensions for base product
-      base_product = ::Registration::SwMgmt.base_product_to_register
+      base_product = if Yast::Mode.update
```
lslezak: Um, we need to be careful that the migration can run also in an installed system. An in that case the mode is very likely *not* set to update. We should check that.

jreidinger: well, at least for now it is not regression as originally only one call was there

PR #457, opened 2019-10-31
Decreasing required memory

schubi2:
Replacing old `Pkg.ResolvableProperties()`and `Pkg.ResolvabeDependencies()` calls by `Y2Packager::Resolvable.any?` and `Y2Packager::Resolvable.find` in order to save memory.

### DIFF
--- a/src/lib/registration/sw_mgmt.rb
+++ b/src/lib/registration/sw_mgmt.rb
@@ -60,10 +60,9 @@ module Registration
 
     ZYPP_DIR = "/etc/zypp".freeze
 
-    FAKE_BASE_PRODUCT = Y2Packager::Resolvable.new( "kind" => :product,
-      "name" => "SLES", "arch" => "x86_64", "version" => "12-0",
+    FAKE_BASE_PRODUCT = { "name" => "SLES", "arch" => "x86_64", "version" => "12-0",
       "flavor" => "DVD", "version_version" => "12", "register_release" => "",
-      "register_target" => "sle-12-x86_64")
+      "register_target" => "sle-12-x86_64" }.freeze
 
     OEM_DIR = "/var/lib/suseRegister/OEM".freeze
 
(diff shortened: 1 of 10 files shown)


src/lib/registration/sw_mgmt.rb
```diff
@@ -653,8 +654,8 @@ def self.select_product_addons(products, addons)
     # find the product resolvables from the specified repository
     def self.products_from_repo(repo_id)
       # TODO: only installed products??
-      Pkg.ResolvableProperties("", :product, "").select do |product|
-        product["source"] == repo_id
+      Y2Packager::Resolvable.find(kind: :product).select do |product|
```
lslezak: You can add more filters to the find call:

```suggestion
      Y2Packager::Resolvable.find(kind: :product, source: repo_id)
```

src/lib/registration/sw_mgmt.rb
```diff
@@ -259,30 +260,29 @@ def self.evaluate_product(p, selected, installed)
...
-      Pkg.ResolvableProperties("", :product, "").any? do |p|
-        p["status"] == :installed || p["status"] == :removed
-      end
+      Y2Packager::Resolvable.any?(kind: :product, status: :installed) ||
+        Y2Packager::Resolvable.any?(kind: :product, status: :removed)
     end
 
     def self.installed_products
       # just for testing/debugging
       return [FAKE_BASE_PRODUCT] if ENV["FAKE_BASE_PRODUCT"]
```
lslezak: The `FAKE_BASE_PRODUCT` is a Hash, this needs to be converted to `Y2Packager::Resolvable` so this always returns the same class.

src/lib/registration/sw_mgmt.rb
```diff
@@ -60,9 +60,11 @@ class SwMgmt
 
     ZYPP_DIR = "/etc/zypp".freeze
 
-    FAKE_BASE_PRODUCT = { "name" => "SLES", "arch" => "x86_64", "version" => "12-0",
+    FAKE_BASE_PRODUCT = { "kind" => :product,
```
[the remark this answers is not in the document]
schubi2: I have change it back to a hash in order not to change the interface of the swmgmt class. 
Otherwise it would break a lot of other stuff in the modules which are using swmgmt.

src/lib/registration/sw_mgmt.rb
```diff
@@ -551,19 +569,19 @@ def self.use_credentials(filename)
...
+      products = Y2Packager::Resolvable.find(kind: :product)
 
       installed_addons = products.select do |product|
-        (product["status"] == :installed || product["status"] == :removed) &&
-          product["type"] != "base"
+        (product.status == :installed || product.status == :removed) &&
+          product.type != "base"
       end
 
-      product_names = installed_addons.map { |a| "#{a["name"]}-#{a["version"]}-#{a["release"]}" }
```
imobachgs: Why are you dropping the "release" part?

schubi2: As far I am remembering Lade has told me that there is no "release" method available. Up to now it has not hurt us because it is a hash.

imobachgs: Ok, fine with me. I just wanted to know the reason. Thanks.

PR #459, opened 2019-11-14
Support base product renames (jsc#SLE-7101)

lslezak:
- Support base product renames when upgrading from the Full medium
- This allows correctly upgrading from SLES11, SLES12 + HPC and other renamed/changed products (see [Y2Packager::ProductUpgrade::MAPPING](https://github.com/yast/yast-yast2/blob/master/library/packages/src/lib/y2packager/product_upgrade.rb#L25))
- Tested manually
- 4.2.17

## Screenshots

Original failure when upgrading from SLES11-SP4:

sles11_upgrade_full_medium_original

With the fix the upgraded product is found and the SLES11-SP4 system can be upgraded to SLES15-SP2:

sles11_upgrade_full_medium_fixed

### DIFF
--- a/src/lib/registration/ui/migration_repos_workflow.rb
+++ b/src/lib/registration/ui/migration_repos_workflow.rb
@@ -618,38 +618,91 @@ module Registration
         Yast::SourceDialogs.SetURL("dvd://")
       end
 
+      # Add and initialize the respective base product repository from the Full medium
+      # for upgrade, select the new base product to install.
       def add_offline_base_product
-        # in offline upgrade add the repository with the selected base product
-        # FIXME: similar to clients/inst_complex_welcome.rb and widgets/product_selector.rb
-        url = Yast::InstURL.installInf2Url("")
-        base_products = Y2Packager::ProductLocation
-                        .scan(url)
-                        .select { |p| p.details && p.details.base }
-        log.info "Found base products on the offline medium: #{base_products.pretty_inspect}"
-
         # find the installed base product
         installed_base = Y2Packager::Resolvable.find(
           kind: :product, status: :installed, type: "base"
         ).first
+
         if !installed_base
           log.error("Installed base product not found")
           return
         end
 
-        # FIXME: handle also the product renames
-        new_base = base_products.find { |p| p.details && p.details.product == installed_base.name }
+        new_base = full_medium_upgrade_product(installed_base)
+
         if !new_base
           log.error("New base product not found")
           return
         end
 
+        new_product = new_base.details && new_base.details.product
+        log.info("Upgrading product #{installed_base.name.inspect} to #{new_product.inspect}")
+
+        add_full_medium_repository(new_product, new_base.dir)
+      end
+
+      #
+      # Scan the installation repository and return the found product directories.
+      #
+      # @return [Array<Y2Packager::ProductLocation>] the found product directories
+      def full_medium_products
+        # in offline upgrade add the repository with the selected base product
+        # FIXME: similar to clients/inst_complex_welcome.rb and widgets/product_selector.rb
+        url = Yast::InstURL.installInf2Url("")
+        products = Y2Packager::ProductLocation
+                   .scan(url)
+                   .select { |p| p.details && p.details.base }
+        log.info("Found base products on the offline medium: #{products.pretty_inspect}")
+        products
+      end
+
+      #
+      # Find the upgrade base product from the Full medium for the installed
+      # base product.
+      #
+      # @param installed_base [<Y2Packager::Resolvable>] the installed base product
+      # @return [Y2Packager::ProductLocation, nil] The location of the new base product on
+      #  the Full installation medium, nil if no suitable upgrade product is found
+      def full_medium_upgrade_product(installed_base)
+        new_base = nil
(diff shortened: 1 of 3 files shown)


src/lib/registration/ui/migration_repos_workflow.rb
```diff
@@ -618,38 +618,91 @@ def prepare_media_upgrade
...
+      # Scan the installation repository and return the found product directories.
+      #
+      # @return [Array<Y2Packager::ProductLocation>] the found product directories
+      def full_medium_products
+        # in offline upgrade add the repository with the selected base product
+        # FIXME: similar to clients/inst_complex_welcome.rb and widgets/product_selector.rb
+        url = Yast::InstURL.installInf2Url("")
+        products = Y2Packager::ProductLocation
+                   .scan(url)
+                   .select { |p| p.details && p.details.base }
```
jreidinger: ah, registration does not use new rubocop, so cannot use `&.`

src/lib/registration/ui/migration_repos_workflow.rb
```diff
@@ -618,38 +618,91 @@ def prepare_media_upgrade
...
         # FIXME: this is the same as in installation/widgets/product_selector.rb and other places
         show_popup = true
+        url = Yast::InstURL.installInf2Url("")
         log_url = Yast::URL.HidePassword(url)
-        Yast::Packages.Initialize_StageInitial(show_popup, url, log_url, new_base.dir)
+        Yast::Packages.Initialize_StageInitial(show_popup, url, log_url, dir)
 
         # select the product to install
-        Yast::Pkg.ResolvableInstall(new_base.details && new_base.details.product, :product, "")
+        Yast::Pkg.ResolvableInstall(new_product, :product, "") if new_product
```
jreidinger: what happens here when nil is passed? Previously you call `new_product = new_base.details && new_base.details.product` so code expect also nil there.

src/lib/registration/ui/migration_repos_workflow.rb
```diff
@@ -618,38 +618,91 @@ def prepare_media_upgrade
...
+        new_base = nil
+        base_products = full_medium_products
+        installed_names = Y2Packager::Resolvable.find(kind: :product, status: :installed)
+                                                .map(&:name)
+        # first check if there is a product rename defined for the installed products
+        # and the new renamed base product is available
+        # key in the mapping: list of installed products, value: the new base product
+        Y2Packager::ProductUpgrade::MAPPING.each do |k, v|
+          if (k - installed_names).empty?
+            new_base = base_products.find { |p| p.details && p.details.product == v }
```
jreidinger: is this `p.details && ` needed? It looks like base_products have to have it defined if I read code correctly.

src/lib/registration/ui/migration_repos_workflow.rb
```diff
@@ -618,38 +618,91 @@ def prepare_media_upgrade
...
+          end
+        end
+
+        # if no product rename was found then use the 1:1 upgrade
+        new_base ||= base_products.find do |p|
+          p.details && p.details.product == installed_base.name
+        end
```
jreidinger: logging here would be really useful.

src/lib/registration/ui/migration_repos_workflow.rb
jreidinger: ah, there is condition. nevermind

src/lib/registration/ui/migration_repos_workflow.rb
lslezak: You are right, it's not strictly required as `nil` should not be there. But I'd rather keep it here to be safe if we move the code somewhere else share it with the other code...

src/lib/registration/ui/migration_repos_workflow.rb
```diff
@@ -618,38 +618,93 @@ def prepare_media_upgrade
...
           return
         end
 
+        new_product = new_base.details && new_base.details.product
+        log.info("Upgrading product #{installed_base.name.inspect} to #{new_product.inspect}")
+
+        add_full_medium_repository(new_product, new_base.dir)
+      end
+
+      #
```
teclator: NP: Any reason for adding it? I have seen it in the next 3 methods.

src/lib/registration/ui/migration_repos_workflow.rb
```diff
@@ -618,38 +618,93 @@ def prepare_media_upgrade
...
+      def full_medium_upgrade_product(installed_base)
+        new_base = nil
+        base_products = full_medium_products
+        installed_names = Y2Packager::Resolvable.find(kind: :product, status: :installed)
+                                                .map(&:name)
+        # first check if there is a product rename defined for the installed products
+        # and the new renamed base product is available
+        # key in the mapping: list of installed products, value: the new base product
+        Y2Packager::ProductUpgrade::MAPPING.each do |k, v|
+          if (k - installed_names).empty?
```
imobachgs: I would have expected RuboCop to complain and ask you to use a guard clause :)

lslezak: Ah, it probably does not complain as we still use the old version here... :wink:

PR #463, opened 2019-12-05
Switch medium type

jreidinger:


src/lib/registration/ui/migration_repos_workflow.rb
```diff
@@ -575,6 +575,8 @@ def system_upgrade_check
         # the system is registered, continue with the SCC/SMT/RMT based upgrade
         elsif Registration.is_registered?
           log.info "The system is registered, using the registration server for upgrade"
+          # act during upgrade for registered system like online even when medium is Full
+          Y2Packager::MediumType.type = :online if Y2Packager::MediumType.offline?
```
lslezak: I'd add a `Stage.initial` check there as well, I'm not sure if this is also used in a migration in the running system. (The media type detection does not work in installed system.)

```suggestion
          Y2Packager::MediumType.type = :online if Stage.initial && Y2Packager::MediumType.offline?
```

PR #465, opened 2019-12-10
Registration detection autoupgrade

jreidinger:
https://trello.com/c/pyqFurT6/1441-3-updater-registration-handling
Goal of this PR is to simplify autoupgrade profile as much as possible. Ideally for registered system empty profile should be enough for upgrade.


src/lib/registration/clients/scc_auto.rb
```diff
@@ -0,0 +1,409 @@
...
+          log.info "Migration result: #{ret}"
+          ret == :next
+        # Full medium we can upgrade without registration
+        elsif Y2Packager::MediumType.offline?
+          true
+        else
+          # Intentionally use blocking popup as it is fatal error that stops installation.
+          Yast::Popup.Error(
+            # TRANSLATORS: profile wants to do registration, but old system is not registered.
+            _("Old system is not registered and autoyast profile require registration.\n" \
```
lslezak: We should improve it when already touching it:

```suggestion
            _("The old system is not registered and the AutoYaST profile requires registration.\n" \
```

PR #466, opened 2020-01-10
Properly remove the old repositories during migration (bsc#1159433)

lslezak:
## The Problem

- https://bugzilla.suse.com/show_bug.cgi?id=1159433
- The old SP1 repositories were not removed from the system when upgrading to SP2

## The Solution

- It turned out that the SCC correctly returned the obsoleted SP1 service for removal, but YaST did not find it in the system so no removal was actually done.
- The problem was a missing repository initialization, the `Pkg.ServiceAliases` returned an empty list for the current services.

## The Fix

- Just (re)load the repositories from the system, then the old services are properly detected and removed.

## Testing

- Tested manually, works fine, see the screenshot below

## Screenshots

There are no SP1 repositories displayed later:

migration_old_repos

*Note: There is still a problem that it offers the new SP2 repositories to remove. But that's another separate problem which will be solved in a separate PR in a different package...*

### DIFF
--- a/src/lib/registration/ui/offline_migration_workflow.rb
+++ b/src/lib/registration/ui/offline_migration_workflow.rb
@@ -77,7 +77,7 @@ module Registration
     private
 
       # force reloading of the old repositories so we can detect and remove the obsoleted
-      # services during migration
+      # services during migration (bsc#1159433)
       def reinit_repos
         Yast::Pkg.SourceFinishAll
         Yast::Pkg.SourceRestore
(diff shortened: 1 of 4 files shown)


src/lib/registration/ui/offline_migration_workflow.rb
```diff
@@ -74,6 +76,13 @@ def main
 
     private
 
+      # force reloading of the old repositories so we can detect and remove the obsoleted
+      # services during migration
```
jreidinger: please add here also bsc number, so we can check in future some details.

PR #467, opened 2020-01-14
[SLE-9109] Packages online search

imobachgs:
This PR is an initial implementation of [the online search feature SLE 9109](https://jira.suse.com/browse/SLE-9109). It is rather incomplete, but it is already able to search for packages and register modules as needed.

online-search-enable-module
*Selecting a package from an unregistered module for installation*

online-search-module-details
*Displaying package details (including version and module where it belongs to)*

online-search-menu-option
*The menu option to start the online search*

## Todo

- [x] Register the base system
- [x] Do not initialize the sw management when called from the packager
- [ ] Install the packages when called as a standalone client
- [ ] Display dependencies when asking about enabling a module
- [x] Refresh module details after selecting a package for installation
- [x] Keep table order after selecting a package

### DIFF
--- a/src/lib/registration/addon.rb
+++ b/src/lib/registration/addon.rb
@@ -28,6 +28,9 @@ require "y2packager/resolvable"
 module Registration
   # this is a wrapper class around SUSE::Connect::Product object
   class Addon
+    extend Yast::I18n
+    include Yast::I18n
+
     # The list of addons has not been loaded
     class AddonsNotLoaded < StandardError; end
 
@@ -350,6 +353,22 @@ module Registration
     # @return [Boolean] true if a not empty EULA url is present; false otherwise
     def eula_acceptance_needed?
       !eula_url.to_s.strip.empty?
+    end
+
+    STATUS_TRANSLATIONS = {
+      registered:    N_("registered"),
+      selected:      N_("to be registered"),
+      auto_selected: N_("to be registered"),
+      available:     N_("not registered yet"),
+      unknown:       N_("unknown")
+    }.freeze
+    private_constant :STATUS_TRANSLATIONS
+
+    # Translates the status into a translated and human readable string
+    #
+    # @return [String]
+    def status_to_human
+      _(STATUS_TRANSLATIONS[status])
     end
 
     def self.dump_addons
--- a/src/lib/registration/addon.rb
+++ b/src/lib/registration/addon.rb
@@ -356,22 +356,6 @@ module Registration
       !eula_url.to_s.strip.empty?
     end
 
-    STATUS_TRANSLATIONS = {
-      registered:    N_("registered"),
-      selected:      N_("to be registered"),
-      auto_selected: N_("to be registered"),
-      available:     N_("not registered yet"),
-      unknown:       N_("unknown")
-    }.freeze
-    private_constant :STATUS_TRANSLATIONS
-
-    # Translates the status into a translated and human readable string
-    #
-    # @return [String]
-    def status_to_human
-      _(STATUS_TRANSLATIONS[status])
-    end
-
     def self.dump_addons
       # dump the downloaded data to a file for easier debugging,
       # avoid write failures when running as an unprivileged user (rspec tests)
--- a/src/lib/registration/addon.rb
+++ b/src/lib/registration/addon.rb
@@ -28,6 +28,9 @@ require "y2packager/resolvable"
 module Registration
   # this is a wrapper class around SUSE::Connect::Product object
   class Addon
+    # The list of addons has not been loaded
+    class AddonsNotLoaded < StandardError; end
+
     class << self
       # read the remote add-on from the registration server
       # @param registration [Registration::Registration] use this object for
@@ -40,6 +43,17 @@ module Registration
         dump_addons
 
         @cached_addons
+      end
+
+      # Find an addon using its ID
+      #
+      # @note This method needs #find_all to be call previously.
+      #
+      # @param id [Integer] Addon's ID
+      # @return [Addon,nil] The addon with the given ID or nil if it was not found
+      def find_by_id(id)
+        raise AddonsNotLoaded unless @cached_addons
+        @cached_addons.find { |a| a.id == id }
       end
 
       def reset!
--- a/src/lib/registration/addon.rb
+++ b/src/lib/registration/addon.rb
@@ -219,6 +219,7 @@ module Registration
     # the constructor
     # @param pure_addon [SUSE::Connect::Product] a pure add-on from the registration server
     def initialize(pure_addon)
+      textdomain "registration"
       @pure_addon = pure_addon
     end
 
--- a/src/lib/registration/clients/online_search.rb
+++ b/src/lib/registration/clients/online_search.rb
@@ -1,4 +1,4 @@
-# Copyright (c) [2019] SUSE LLC
+# Copyright (c) [2020] SUSE LLC
 #
 # All Rights Reserved.
 #
--- a/src/lib/registration/clients/online_search.rb
+++ b/src/lib/registration/clients/online_search.rb
@@ -1,0 +1,140 @@
+# Copyright (c) [2019] SUSE LLC
+#
+# All Rights Reserved.
+#
+# This program is free software; you can redistribute it and/or modify it
+# under the terms of version 2 of the GNU General Public License as published
+# by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+# FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License for
+# more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, contact SUSE LLC.
+#
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+
+require "yast"
+require "registration/dialogs/online_search"
+require "registration/addon"
+require "registration/registration"
+require "registration/registration_ui"
+require "registration/sw_mgmt"
+require "registration/url_helpers"
+
+Yast.import "Installation"
+Yast.import "Pkg"
+Yast.import "Sequencer"
+
+module Registration
+  module Clients
+    # Online search client
+    #
+    # It offers an interface to search for packages even if they are included in
+    # modules which has not been activated yet. It is that the case, the user
+    # can select the module for registration before trying to install the package.
+    #
+    # This client will enable any needed module but package installation should
+    # be performed by the regular package manager (`sw_single` client).
+    #
+    # @see Dialogs::OnlineSearch
+    class OnlineSearch
+      include Yast::I18n
+      extend Yast::I18n
+      include Yast::Logger
+
+      def initialize
+        textdomain "registration"
+      end
+
+      # Sequence workflow aliases
+      #
+      # @see #find_addons
+      # @see #package_search
+      # @see #commit
+      def workflow_aliases
+        {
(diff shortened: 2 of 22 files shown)


lslezak:
I haven't checked the code yet, here are just some notes and hints for improvements :wink:

### UI/UX

- It's not clear what can be entered in the search field, is it just a substring or regexps are supported? The full packager widget contains a `Search Mode` combo box with values like "Contains", "Begins with", or "Use Regular Expression". You can also search in description or in the package summary. I think it would be nice to use it also here if the SCC backend supports that...
- About the search: Would it be possible to filter out e.g. the paid extensions? (The ones which require a separate registration key.)
- Does the API return some more details about the package? E.g. summary or details? Sometimes it might be difficult to just decide from the package name.
- When a new product to register requires a registration key the user should be informed in advance, before really registering it.
- The package and product details are mixed together in the `RichText` details at the bottom. You can use e.g. `<h3>`  headers with "Package" and "Product" label. Or at least there should be an empty line between the package and the product details.
- Do we need that `Toggle Package` button at all? The standard packager widget does not use anything like that...
- I'm not sure if we need a separate "Product Status" column. Maybe simply appending  something like "(Not registered yet)" at the end would be better. E.g. display "Containers Module (Not registered yet)" in the "Product" column.


### Technical Questions

- Will this also later work with SMT/RMT? If not (or at least not in the near future) we should check which registration server is actually used and display an informative error for SMT/RMT.

### Todo Notes

> Return the list of packages that were selected for installation.

I think that you could simply select the packages to install after adding the respective modules, the packager widget would then correctly display them as selected to install.

src/lib/registration/remote_package.rb
```diff
@@ -0,0 +1,85 @@
...
+# find current contact information at www.suse.com.
+
+module Registration
+  # Represents a package as seen through the SUSE::Connect API
+  #
+  # Additionally, it offers a mechanism to find out the local package (libzypp counterpart).
+  #
+  # @example Find the status
+  class RemotePackage
+    attr_reader :name, :arch, :version, :release, :addon
```
lslezak: We already have some classes dealing with `name`,  `version`, `arch` in YaST. Would it be possible to reuse them here? Or at least use as a base class?

If not then it seems we should probably create a base class for all those uses and share it...

src/lib/registration/widgets/remote_package_details.rb
```diff
@@ -0,0 +1,72 @@
...
+        textdomain("registration")
+        super
+      end
+
+      # Updates the widget's content
+      #
+      # @param package [RemotePackage] Package obtained via online search
+      def update(package)
+        lines = [
+          format(_("<b>Name:</b> %{package_name}"), package_name: package.name),
```
lslezak: All values should be HTML escaped. Although it's almost sure that e.g. `<` will not be part of the package name we should be on the safe side...

Use `h()` from `ERB::Util` like [here](https://github.com/yast/yast-registration/blob/master/src/data/registration/certificate_summary.erb#L7).

This looks quite trivial so we could built the text inline here, for more complicated cases we could use an external ERB template ([like we already do](https://github.com/yast/yast-registration/tree/master/src/data/registration)).

src/lib/registration/addon.rb
```diff
@@ -338,6 +356,22 @@ def eula_acceptance_needed?
       !eula_url.to_s.strip.empty?
     end
 
+    STATUS_TRANSLATIONS = {
```
lslezak: The name is a bit confusing, these are actually not translations, `N_()` actually returns the original string. Maybe `STATUS_NAMES` or something like that would be more clear.

src/lib/registration/addon.rb
```diff
@@ -338,6 +356,22 @@ def eula_acceptance_needed?
...
+      auto_selected: N_("to be registered"),
+      available:     N_("not registered yet"),
+      unknown:       N_("unknown")
+    }.freeze
+    private_constant :STATUS_TRANSLATIONS
+
+    # Translates the status into a translated and human readable string
+    #
+    # @return [String]
+    def status_to_human
```
lslezak: Something like `status_to_text` sounds better to me...

src/lib/registration/clients/online_search.rb
```diff
@@ -0,0 +1,176 @@
+# Copyright (c) [2019] SUSE LLC
```
lslezak: It's already 2020... :wink: 
```suggestion
# Copyright (c) [2020] SUSE LLC
```

src/lib/registration/dialogs/online_search.rb
```diff
@@ -0,0 +1,73 @@
...
+      # Constructor
+      def initialize
+        textdomain "registration"
+        @selected_packages = []
+        super
+      end
+
+      # @macro seeAbstractWidget
+      def title
+        _("Online Search")
```
lslezak: Maybe a `TRANSLATORS:`  comment could be useful here...

src/lib/registration/widgets/package_search.rb
```diff
@@ -0,0 +1,215 @@
+# Copyright (c) [2019] SUSE LLC
```
lslezak:
```suggestion
# Copyright (c) [2020] SUSE LLC
```

src/lib/registration/dialogs/online_search.rb
```diff
@@ -0,0 +1,73 @@
+# Copyright (c) [2019] SUSE LLC
```
lslezak:
```suggestion
# Copyright (c) [2020] SUSE LLC
```

src/lib/registration/package_search.rb
```diff
@@ -0,0 +1,100 @@
+# Copyright (c) [2019] SUSE LLC
```
lslezak:
```suggestion
# Copyright (c) [2020] SUSE LLC
```

src/lib/registration/widgets/package_search.rb
```diff
@@ -0,0 +1,215 @@
...
+        event["WidgetID"] == "search_form_button" ||
+          (event["WidgetID"] == "search_form_text" && event["EventReason"] == "Activated")
+      end
+
+      # Performs the search and updates the packages table
+      #
+      # @param text [String] Text to search for
+      def search_package(text)
+        @search = ::Registration::PackageSearch.new(text: text)
+        Yast::Popup.Feedback(_("Searching..."), _("Searching for packages with the given name")) do
```
lslezak: `TRANSLATORS:` comments?

BTW I would use something more generic instead of the `with the given name` part. Just in case we later allow searching in summary or package description...

src/lib/registration/widgets/package_search.rb
```diff
@@ -0,0 +1,215 @@
...
+      end
+
+      # Asks the user to enable the addon
+      #
+      # It omits the question if the addon is already registered or selected for registration.
+      #
+      # @param addon [Addon] Addon to ask about
+      def enable_addon?(addon)
+        message = format(
+          _("'%{name}' module is not enabled for this system.\nDo you want to enable it?"),
```
lslezak: There are modules and extensions

```suggestion
          _("'%{name}' module/extension is not enabled for this system.\nDo you want to enable it?"),
```

would be probably more correct.

src/lib/registration/widgets/package_search_form.rb
```diff
@@ -0,0 +1,63 @@
+# Copyright (c) [2019] SUSE LLC
```
lslezak:
```suggestion
# Copyright (c) [2020] SUSE LLC
```

src/lib/registration/widgets/package_search_form.rb
```diff
@@ -0,0 +1,63 @@
...
+    # start the query.
+    class PackageSearchForm < ::CWM::CustomWidget
+      def initialize
+        textdomain "registration"
+        super
+      end
+
+      def contents
+        Frame(
+          _("Search"),
```
lslezak: Unfortunately YaST does not support gettext contexts, but it would be nice to write some TRANSLATORS: comments for the labels and push buttons.

src/lib/registration/widgets/remote_package_details.rb
```diff
@@ -0,0 +1,71 @@
+# Copyright (c) [2019] SUSE LLC
```
lslezak:
```suggestion
# Copyright (c) [2020] SUSE LLC
```

src/lib/registration/widgets/remote_package_details.rb
```diff
@@ -0,0 +1,71 @@
...
+        super
+      end
+
+      # Updates the widget's content
+      #
+      # @param package [RemotePackage] Package obtained via online search
+      def update(package)
+        lines = [
+          format(_("<b>Name:</b> %{package_name}"), package_name: ERB::Util.h(package.name)),
+          format(_("<b>Version:</b> %{package_version}"), package_version: package.full_version),
```
lslezak: NP: I'd use `ERB::Util.h` everywhere, just to be consistent. And if somebody does copy&paste and use a differnt variable it will be still safe...

src/lib/registration/widgets/remote_package_details.rb
```diff
@@ -0,0 +1,71 @@
...
+
+    private
+
+      # Displays the status of the given addon
+      #
+      # @param addon [Addon] Addon to display the status for
+      # @return [String]
+      def addon_status(addon)
+        if addon.registered?
+          _("registered")
```
lslezak: `TRANSLATORS:` comments here are a must, it would be really difficult to translate these texts right without any context...

src/lib/registration/widgets/remote_packages_table.rb
```diff
@@ -0,0 +1,94 @@
+# Copyright (c) [2019] SUSE LLC
```
lslezak:
```suggestion
# Copyright (c) [2020] SUSE LLC
```

src/lib/registration/widgets/remote_packages_table.rb
```diff
@@ -0,0 +1,94 @@
...
+      end
+
+      # @macro seeAbstractWidget
+      def opt
+        [:notify, :immediate, :keepSorting]
+      end
+
+      # @macro seeTable
+      def header
+        [Center(_("Status")), _("Name"), _("Product")]
```
lslezak: `TRANSLATORS:` ...

src/lib/registration/widgets/remote_packages_table.rb
```diff
@@ -0,0 +1,94 @@
...
+      # @param item [RemotePackage] Package to update
+      def update_item(item)
+        columns_for_item(item).each_with_index do |content, idx|
+          change_cell(Id(item.name), idx, content)
+        end
+      end
+
+    private
+
+      # @see http://www.rubydoc.info/github/yast/yast-yast2/CWM%2FTable
```
lslezak: NP
```suggestion
      # @see https://www.rubydoc.info/github/yast/yast-yast2/CWM%2FTable
```

test/registration/package_search_test.rb
```diff
@@ -0,0 +1,137 @@
+# Copyright (c) [2019] SUSE LLC
```
lslezak:
```suggestion
# Copyright (c) [2020] SUSE LLC
```

src/lib/registration/remote_package.rb
imobachgs: Having a class to handle versioning would be nice, but I would postpone it by now. We are investing too much time in this feature and I prefer to release as soon as possible.

src/lib/registration/clients/online_search.rb
imobachgs: Hehe, you are right, time flies :smile:. I started to implement this feature in December 2019, hence the reason to have outdated headers... I will fix them.

PR #468, opened 2020-01-14
Fix scc auto

jreidinger:
*(Description copied from https://github.com/yast/yast-registration/pull/469 :wink:)*

Reported as https://bugzilla.suse.com/show_bug.cgi?id=1160909

Looking at the code, it's obvious that `Registration` was used instead of `::Registration`.

### DIFF
--- a/package/yast2-registration.changes
+++ b/package/yast2-registration.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Tue Jan 14 15:04:15 UTC 2020 - Josef Reidinger <jreidinger@suse.com>
+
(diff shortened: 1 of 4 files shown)


ancorgs:
I also created https://github.com/yast/yast-registration/pull/469, but looks you were faster. I will close mine.

ancorgs:
On the other hand, my PR (already closed) was green in Travis and contained a description. :wink:

jreidinger:
@ancorgs on other hand this one contain unit tests that was red and now is green. Even not perfect as we have https://github.com/yast/yast-registration/blob/master/src/lib/registration/connect_helpers.rb#L146 which blocks all exceptions

PR #469, opened 2020-01-14
Fix crash in AutoYaST when connecting to SCC (bsc#1160909)

ancorgs:
Reported as https://bugzilla.suse.com/show_bug.cgi?id=1160909

Looking at the code, it's obvious that `Registration` was used instead of `::Registration`.


ancorgs:
closed in favor of #468

PR #470, opened 2020-01-15
AY: Providing a "retry" button if a request timed out.

schubi2:
**Problem**
If registration is needed while an AutoYaST installation the installation will be broken if there is
network is not available or it is not stable.

**Solution**
Telling the user that there is a problem with the network and giving him a chance to retry the
connection in order to go on with the installation.


lslezak:
Um, does it actually make sense to have that `Retry` button in AutoYaST? Sure, that's better than a crash, but in the AutoYaST mode I'd expect some auto-retry. E.g. automatically retry after a minute, do like 3 attempts. If that still fails display the popup for the user.

PR #476, opened 2020-01-31
[JSE-9109] Add a package search controller

imobachgs:
This PRs extracts business logic from `Widgets::PackageSearch` to a separate `Controllers::PackageSearch` class. This class will be responsible for keeping the state and implementing the actions of the package search feature.

I am keeping the same version number because (hopefully) the user will not notice any difference.

### DIFF
--- a/src/lib/registration/controllers/package_search.rb
+++ b/src/lib/registration/controllers/package_search.rb
@@ -88,10 +88,13 @@ module Registration
 
       # Selects the given addon if needed
       #
+      # If the addon is registered or selected, does nothing. If the addon
+      # was auto selected, it will be marked as selected.
+      #
       # @param addon [Addon] Addon to select
       def select_addon(addon)
-        return if addon.registered? || addon.selected? || addon.auto_selected?
-        addon.selected if enable_addon?(addon)
+        return if addon.registered? || addon.selected?
+        addon.selected if addon.auto_selected? || enable_addon?(addon)
       end
 
       # Unselects the given addon if required
--- a/src/lib/registration/controllers/package_search.rb
+++ b/src/lib/registration/controllers/package_search.rb
@@ -38,22 +38,13 @@ module Registration
         @selected_packages = []
       end
 
-      # Returns the list of the current search
-      #
-      # @return [Array<RemotePackage>] List of found packages
-      def packages
-        @search ? @search.packages : []
-      end
-
       # Performs a package search
       #
       # @param text [String] Term to search for
+      # @return [Array<Registration::RemotePackage>] List of packages
       def search(text)
         @search = ::Registration::PackageSearch.new(text: text)
-        selected_package_ids = selected_packages.map(&:id)
-        @search.packages.each do |pkg|
-          pkg.select! if selected_package_ids.include?(pkg.id)
-        end
+        @search.packages
       end
 
       # Selects/unselects the current package for installation
--- a/src/lib/registration/controllers/package_search.rb
+++ b/src/lib/registration/controllers/package_search.rb
@@ -189,8 +189,8 @@ module Registration
       # @param msg [String] Message to display at the beginning of the line
       # @param addon [Registration::Addon]
       def log_addon(msg, addon)
-        log.info "#{msg}: #{addon.inspect}, registered=#{addon.registered?}, selected=#{addon.selected?}" \
-          "auto_selected=#{addon.auto_selected?}"
+        log.info "#{msg}: #{addon.inspect}, registered=#{addon.registered?}, " \
+          "selected=#{addon.selected?}, auto_selected=#{addon.auto_selected?}"
       end
     end
   end
--- a/src/lib/registration/controllers/package_search.rb
+++ b/src/lib/registration/controllers/package_search.rb
@@ -21,11 +21,23 @@ require "yast"
 require "registration/package_search"
 
 Yast.import "Popup"
+Yast.import "HTML"
 
 module Registration
   module Controllers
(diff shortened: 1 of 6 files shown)


src/lib/registration/controllers/package_search.rb
```diff
@@ -0,0 +1,189 @@
...
+      # Constructor
+      def initialize
+        textdomain "registration"
+        @selected_packages = []
+      end
+
+      # Returns the list of the current search
+      #
+      # @return [Array<RemotePackage>] List of found packages
+      def packages
```
jreidinger: I am lost here. What is difference to selected packages? This is all packages that are result of search even not selected?

src/lib/registration/controllers/package_search.rb
```diff
@@ -0,0 +1,189 @@
...
+        @search ? @search.packages : []
+      end
+
+      # Performs a package search
+      #
+      # @param text [String] Term to search for
+      def search(text)
```
jreidinger: To be honest I do not like this search with such huge side effect. and also with strict call order, as packages depends on search called in advance.
In general I do not like much mixing of holder for selection together with search filtering. I would keep it separate.

src/lib/registration/widgets/package_search.rb
```diff
@@ -141,151 +139,39 @@ def start_search_event?(event)
...
       def search_package(text)
         return unless valid_search_text?(text)
-        @search = ::Registration::PackageSearch.new(text: text)
         # TRANSLATORS: searching for packages
         Yast::Popup.Feedback(_("Searching..."), _("Searching for packages")) do
-          selected_package_ids = selected_packages.map(&:id)
-          @search.packages.each do |pkg|
-            pkg.select! if selected_package_ids.include?(pkg.id)
-          end
+          controller.search(text)
```
jreidinger: I expect here call to search as it is not affected by controller and then mark selected packages from controller.

src/lib/registration/widgets/package_search.rb
```diff
@@ -141,151 +139,39 @@ def start_search_event?(event)
...
       def toggle_package
         package = find_current_package
-        return if package.nil? || package.installed?
-
-        if package.selected?
-          unselect_package(package)
-        else
-          select_package(package)
-        end
+        return if package.nil?
```
jreidinger: this should not happen, not? If toggle is disabled when nothing is selected, it should not be possible to get it.

imobachgs:
As discussed on IRC, I will move the list of packages from the current search to the widget, keeping the selected packages to the controller. Thanks for the feedback!

src/lib/registration/controllers/package_search.rb
imobachgs: `#packages` is the list of found packages, and `#selected_packages` is the current selection of packages. Now that we agreed to move the `#packages` list in the UI it should not be so confusing.

src/lib/registration/controllers/package_search.rb
```diff
@@ -0,0 +1,180 @@
...
+
+      # Selects the current package for installation
+      #
+      # If required, it selects the corresponding addon
+      #
+      # @param package [RemotePackage] Package to select
+      def select_package(package)
+        addon = package.addon
+        select_addon(addon) if addon
+        set_package_as_selected(package) if addon.nil? || addon.selected? || addon.registered?
```
jreidinger: NP: this condition looks a bit duplicate and complex. What do you think about having select_addon that return boolean ( now only partially) that looks something like:
```
def select_addon(addon)
    return true if addon.nil? || addon.registered? || addon.selected? || addon.auto_selected?
￼    addon.selected if enable_addon?(addon)
￼end
```

Also that brings me to another question. What happen here if addon is just auto_selected? it looks like it skips setting package as selected, not?

src/lib/registration/controllers/package_search.rb
```diff
@@ -0,0 +1,180 @@
...
+        return if addon.registered? || addon.selected? || addon.auto_selected?
+        addon.selected if enable_addon?(addon)
+      end
+
+      # Unselects the given addon if required
+      #
+      # @param addon [Addon] Addon to unselect
+      def unselect_addon(addon)
+        return if addon.registered? || needed_addon?(addon)
+        addon.unselected if disable_addon?(addon)
```
jreidinger: in general as I see there a lot of conditions, I miss a bit more logging there to see what happen and why. Especially if issue appears on customer machine.

src/lib/registration/controllers/package_search.rb
```diff
@@ -0,0 +1,180 @@
...
+        yes_no_popup(description + question)
+      end
+
+      # Asks the user to disable the addon
+      #
+      # @param addon [Addon] Addon to ask about
+      def disable_addon?(addon)
+        message = format(
+          _("The '%{name}' is not needed anymore.\n" \
+            "Do you want to unselect it?"),
```
jreidinger: just curious does it deselect auto dependencies that is no longer needed?

src/lib/registration/controllers/package_search.rb
imobachgs: Actually, that was a bug. If an addon is marked as '#auto_selected', it should be marked as '#selected' so it does not get removed accidentally. Fixed now.

src/lib/registration/controllers/package_search.rb
imobachgs: Yes, it does. We are not informing the user (we need to do some calculation). However, the overall mechanism to inform the user (and the workflow in general) are still under discussion. I would like to keep the current behavior and fix/improve it later.

PR #483, opened 2020-02-27
Refactor addons selection dialog using a new widget

dgdavid:
### TL;DR

Implements a new dialog to manage the addons selection, relying in a new custom widget designed to display the selection in a master-detail interface.

This widget fixed the problems with the scroll and _selection_ in yast2-registration without the `MultiSelectionBox` limitations.

Related to 
  - https://trello.com/c/CmaNrDAw
  - https://bugzilla.suse.com/show_bug.cgi?id=1049965
  - https://bugzilla.suse.com/show_bug.cgi?id=1161200

<details>
<summary>Click to show/hide a short video with those changes during installation</summary>

---

ezgif com-optimize(4)

</details>

### The long story

Currently, we have at least two dialogs[1][2] following the same pattern: a multi-selection list on top and an area to display more details about a specific item at the bottom. A [master-detail interface](https://en.wikipedia.org/wiki/Master%E2%80%93detail_interface).

Both of them deal with a multi-status selection since an addon or module extension can be (un)selected by the user or auto-selected as a dependency, but following different approaches: [`yast2-registration` uses a RichText widget to simulate the checkboxes](https://github.com/yast/yast-registration/blob/9c04bc3d0c44488425f056f781d0c091a9efb10e/src/lib/registration/ui/addon_selection_base_dialog.rb#L195-L199) through pseudo-HTML links (`<a>` tag)  while [`yast2-packager` builds the selector using the `MultiSelectionBox` widget](https://github.com/yast/yast-packager/blob/6c0cd4cb4b46bf64e836d910f253941ff0eebd5f/src/lib/y2packager/dialogs/addon_selector.rb#L132-L137). However, both implementations have issues

* yast2-registration/RichText trick
  - reset the scroll, when present, after each selection
  - force to select/unselect (or vice-versa) the item to see its details
* yast2-packager/MultiSelectionBox
  - do not have support to represent the `auto_selected` status
  - do not allow to cycle manually from auto-selected to selected status
  - force the select/unselect behavior to display the details too

We tried to make use fo the recently created `CustomStatusMultiItemSelector` widget, but its current implementation does not fulfill the requirements to solve those issues:

* it requires an inline description, which is not desired for this use case because 
  - an addon description is long enough and can contain rich text, which is not correctly displayed in text mode with that widget
  - the vertical scroll will be too long
  - the widget does not perform an automatic text wrapping, forcing to cut the description lines at an arbitrary line width
* it does not allow to click over one item without selecting it

So, after a few meetings, we decided to go ahead with a workaround for the RichText approach using two pseudo-HTML links for each item to simulate both, the manual (un)selection action and the single click over the item without selecting it.

#### New solution, different _issues_

In the GUI mode, where the user mostly will use the mouse to interact with the installer, the workaround works as well as it would be expected. But in text mode still being room for improvements, namely

- The RichText microfocus: [a problem already described some time ago](https://gist.github.com/dgdavid/926b8cb93c79d2ec3720758266d4bf31): when restoring the scroll position, the first link visible will catch focus (actually, the RichText Microfocus). Ideally, in order to achieve a better simulation, the link originating the _redraw_ should catch it instead. However, we cannot forget that we are putting the RichText to the limit, using it as a workaround to make a work for what it was not designed.

- It would be nice to handle the key up/down strokes in the same way that [is already done with space and return keys](https://github.com/libyui/libyui-ncurses/blob/f9a6b1c01a4b3e681ec79126fb32e2dd5162982a/src/NCRichText.cc#L270-L296): firing and event when a link catches the microfocus. Doing that, the widget will not need to split each item into two links, avoiding to have a _weird navigation_ and, especially, allowing to react to this event and udpate the item details as soon as the item gets that microfocus. I.e, allowing to keep the details in sync with the master navigation.

<details>
<summary>Click to show/hide a short video to see how it works in an 80x25</summary>

---

ezgif com-add-text

</details>

---

[1] https://github.com/yast/yast-registration/blob/9c04bc3d0c44488425f056f781d0c091a9efb10e/src/lib/registration/ui/addon_selection_base_dialog.rb
[2] https://github.com/yast/yast-packager/blob/6c0cd4cb4b46bf64e836d910f253941ff0eebd5f/src/lib/y2packager/dialogs/addon_selector.rb


src/lib/registration/controllers/addons_selection.rb
```diff
@@ -0,0 +1,259 @@
...
+      end
+
+      # Updates the filter value
+      #
+      # @param enabled [Boolean] the new value for the filter
+      def filter=(enabled)
+        self.class.filtering_unreleased = enabled
+      end
+
+      # FIXME: still relevant?
```
dgdavid: I'd like to know if this piece of code still being relevant. I have ever seen that screen with two columns to introduce registration codes.

Could you please take a look and give me a clue?

Thanks!

src/lib/registration/controllers/addons_selection.rb
```diff
@@ -0,0 +1,259 @@
...
+          item_id_for(addon),
+          addon.label,
+          addon.status,
+          "<h3>#{addon.friendly_name}</h3>#{addon.description}"
+        )
+      end
+
+      # Returns the available addons based in the filter status
+      #
+      # When filtering, developments addons will be rejected unless the are selected or
```
dgdavid: s/developments/development

src/lib/registration/dialogs/addons_selection.rb
```diff
@@ -0,0 +1,135 @@
...
+
+      # @return [Array<Symbol>]
+      def continue_actions
+        @continue_actions ||= [:next, :back, :cancel, :abort, :skip]
+      end
+
+      # @macro seeDialog
+      def help
+        [
+          _("<p>Here you can select available extensions and modules for your system.</p>"),
```
dgdavid: I tried to keep the translation as untouched as possible, but I has to change some of them a little bit. Shall we simply change them as needed?

src/lib/registration/dialogs/addons_selection.rb
```diff
@@ -0,0 +1,135 @@
...
+      # @macro seeDialog
+      def contents
+        VBox(
+          Left(Heading(_("Available Extensions and Modules"))),
+          addons_selector_widget
+        )
+      end
+
+      # @macro seeDialog
+      def run
```
dgdavid: What about skipping the dialog when there are not available addons?

src/lib/registration/widgets/checkbox_item.rb
```diff
@@ -0,0 +1,191 @@
...
+# find current contact information at www.suse.com.
+
+require "yast"
+
+module Registration
+  module Widgets
+    # A plain Ruby object in charge to build an item "checkbox" representation to be used by
+    # {MasterDetailSelector} widget in a RichText widget.
+    #
+    # FIXME: give support for disabled items; see #icon
```
dgdavid: Question, does exist the possibility to have _disabled_ addons? If so, when? If not, I think we could go ahead by now and add support for this later. What do you think?

src/lib/registration/widgets/checkbox_item.rb
```diff
@@ -0,0 +1,191 @@
...
+
+      # Path to the icons in the system
+      IMAGES_DIR = "/usr/share/YaST2/theme/current/wizard".freeze
+      private_constant :IMAGES_DIR
+
+      # Help text
+      #
+      # @return [String]
+      def self.help
+        # TRANSLATORS: help text for checked check box
```
dgdavid: Note, the widget should be agnostic enough. That's why I changed the translations for the checkbox legend.

src/lib/registration/widgets/master_detail_selector.rb
```diff
@@ -0,0 +1,176 @@
...
+
+    private
+
+      attr_reader :controller
+
+      # Widget representing the filter when needed
+      #
+      # @return [Yast::Term] CheckBox if the filter should be included; Empty otherwise
+      def filter
+        return Empty() unless controller.include_filter?
```
dgdavid: Although it works, I'm not completely happy on how the filter is being _injected_. A better idea is very welcome.

src/clients/test_addon_selection.rb
```diff
@@ -0,0 +1,2 @@
+require "registration/clients/test_addon_selection"
+Yast::TestAddonSelectionClient.new.main
```
jreidinger: I think this should not get into production.

src/lib/registration/controllers/addons_selection.rb
```diff
@@ -0,0 +1,259 @@
...
+  module Controllers
+    # Controller to manage actions and state of addon selection
+    #
+    # Needed by {Widgets::MasterDetailSelector}
+    class AddonsSelection
+      include Yast::I18n
+
+      class << self
+        # @return [Boolean] If not released (e.g., alpha or beta versions) addons should be filtered
+        attr_accessor :filtering_unreleased
```
jreidinger: I would call it like `release_only`

src/lib/registration/controllers/addons_selection.rb
```diff
@@ -0,0 +1,259 @@
...
+      end
+
+      # A checkbox item representation, see {Widgets::CheckboxItem}
+      Item = Struct.new(:id, :label, :status, :description)
```
jreidinger: This is a bit confusing as Item is also in Yast::UIShortcuts. I would choose a different constant name.

src/lib/registration/widgets/checkbox_item.rb
```diff
@@ -0,0 +1,191 @@
...
+
+        help_text.join
+      end
+
+      # Returns the icon to be used for an item with given status
+      #
+      # @see #icon
+      #
+      # @return [String]
+      def self.icon_for(status)
```
jreidinger: this looks very strange. I expect to get it without instance.

src/lib/registration/controllers/addons_selection.rb
dgdavid: Thanks for the info @lslezak 

Maybe I'm wrong, but I'd say that I saw that _registration code request_ in a single dialog per addon. That's why I asked if the code still relevant.

Of course, keeping the code there does not hurt.

src/lib/registration/dialogs/addons_selection.rb
dgdavid: So, maybe its better to use a "single" translation. Right?

src/lib/registration/widgets/checkbox_item.rb
dgdavid: Again, thanks.

Now I have a better picture of this situation. I was requesting to postpone it because I thought that was a use case not longer available. Indeed, I think supporting _disabled_ items it's a nice to have for the widget. Now I know that actually it's a must.

src/lib/registration/dialogs/addons_selection.rb
lslezak: Well, it depends. One big translation has a disadvantage that if you change just one word all  translations ares invalidated. With smaller parts only that smaller parts needs to be updated by translators.

And as you already wrote "I has to change some of them a little bit" it seems that having more smaller parts is better here.

src/lib/registration/widgets/checkbox_list.rb
```diff
@@ -0,0 +1,70 @@
...
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+
+require "yast"
+require "cwm/common_widgets"
+
+module Registration
+  module Widgets
+    # A check boxes list intended to be the master widget in {MasterDetailSelector}
+    class CheckboxList < CWM::RichText
```
jreidinger: I do not like this class at all :) It pretends to be checkbox list, but it is in fact just richtext box which remember scrolling position.

src/lib/registration/widgets/item_details.rb
```diff
@@ -0,0 +1,63 @@
...
+require "cwm/common_widgets"
+
+module Registration
+  module Widgets
+    # A CWM::RictText intended to be the widget to display item details in {MasterDetailSelector}
+    class ItemDetails < CWM::RichText
+      # Constructor
+      #
+      # @param placeholder [String] the placeholder to be shown as initial content
+      def initialize(placeholder: nil)
```
jreidinger: it mentions string as possible value, but default is not string :) In general why not having `placeholder: ""` ?

src/lib/registration/widgets/item_details.rb
```diff
@@ -0,0 +1,63 @@
...
+      # Updates the widget's content
+      #
+      # @param content [String] the content to be displayed
+      def update(content)
+        self.value = content
+      end
+
+      # Disables  the widget and resets its content
+      def reset
+        disable
```
jreidinger: why it is needed?

src/lib/registration/widgets/master_detail_selector.rb
jreidinger: make filter own CWM widget that can detect itself how to represent itself.

src/lib/registration/widgets/item_details.rb
dgdavid: Because originally that _detail area_ was disabled. I have no problem getting rid of this behavior and keeping that area always enabled

src/lib/registration/widgets/checkbox_item.rb
dgdavid: Done, see https://github.com/yast/yast-registration/pull/483/commits/b8a381b127c9b54746995c63ca9e077e745304c6

src/lib/registration/widgets/item_details.rb
dgdavid: A leftover, thanks for catching

src/lib/registration/widgets/checkbox_item.rb
dgdavid: Fixed :) See https://github.com/yast/yast-registration/pull/483/commits/7488ee247e0d468c2139aac14675947577c4ef36

src/lib/registration/widgets/checkbox_list.rb
dgdavid: Yes, this a convenience and quite badly named widget. It's here to avoid adding the logic to update (`Yast::UI.ChangeWidget(Id(...), :Value, new_content)`) and remember the scrolling of the _mater area_ directly from the `MasterDetailSelector` widget.

But maybe we should simply move it there, although I must admit that I don't like too much the idea. 

Could you please help me to find a better solution? (taking into account that we are looking for a reusable multi-state selector widget).

I will be extremely grateful to you for doing so.

jreidinger: OK, so what responsibility/features we need here?

1. remember scrolling area - This should in my opinion should go to CWM::RichText itself as flag and feature as other richtext area can benefit from it. ( so basically if flag is set modify value= method ).

2. checkboxes representation in Richtext - it is now in Checkbox Item. It looks fine for me. But what I do not like is that this widget is stupid and get just content. I think it should get list of Checkbox Item and render it itself like e.g. Tabs or Trees do it.

3. reaction to click so it changes states ( like default action that button will changed to selected/unselected ). I think this definitively should be handled here. As part of it I see also method that recognize if event belongs to widget.

4. updating whole widget as there is auto dependencies between addons. This for me belongs to non-generic AddonsList which is child of CheckBox list and which has own handle that do this redrawing and recomputing. This is usual CWM way. You are building bricks of widgets together and us the widget that fits you best.

dgdavid: First of all, thanks for the detailed answer @jreidinger, but I fear I do not completely agree with that vision

Probably I should have started commenting that those widgets are stupid on purpose. They are there only to "represent" the information, not to deal with it. They have no other responsibility. They know nothing about the selection logic or selected items. They have no state.

That's the way I found more appropriate to make them reusable. And that's why the third player, the controller, is needed. It is its responsibility to provide the information and deal with it.

Then, in this PR we have

* A main/wrapper/container widget (aka `MasterDetailSelector`): it has no state either, it's the only responsibility is react to certain events, requesting needed actions to the controller and keeping its child widgets updated and in sync.
* The list container (aka `CheckboxList`): will contain all checkboxes items and provides the single #update method.
* The detail container (aka `ItemDetails`): mostly the same as the list, but for the details of the currently selected item.
* The controller: is the key piece here. It knows where to look for the information, how to make the selection and auto-selection, etc. It should provide all necessary information and actions needed by the main widget.

Of course, those two child widgets are only convenience widgets, we can do the same directly from the `MasterDetailSelector` but I thought it would be a good idea to keep the main widget agnostic enough: in the end, it only needs to call  `#update` or `#reset` method. The rest of the story (saving/restoring the scroll, the place holder, etc) is not its business.

I don't know if I explained it better now or if that philosophy does not fit with CWM, but that's the approach I wanted to follow: making a _representational widgets_ that only knows how to present the information to the user and how to ask for it. Nothing more.

Do you still thinking the widgets need to be stateful? If so, are you convinced they still being reusable for other use cases?

Apart from those questions, do you think we could at least give it a try in `yast2-registration` before refining them and move it to a more centric place?

Thanks.

dgdavid: > This should in my opinion should go to CWM::RichText itself as flag and feature as other richtext area can benefit from it

That's a good idea. So, if I got it correctly, we should fix it [over here](https://github.com/yast/yast-yast2/blob/b0eebf01c5485448393933563082cb908de0d3fc/library/cwm/src/lib/cwm/common_widgets.rb#L364-L368) make the PR and then update the [yast2 dependency](https://github.com/yast/yast-registration/blob/9c04bc3d0c44488425f056f781d0c091a9efb10e/package/yast2-registration.spec#L43) here, right?

jreidinger: > > This should in my opinion should go to CWM::RichText itself as flag and feature as other richtext area can benefit from it
> 
> That's a good idea. So, if I got it correctly, we should fix it in cwm/common_widgets ([over here](https://github.com/yast/yast-yast2/blob/b0eebf01c5485448393933563082cb908de0d3fc/library/cwm/src/lib/cwm/common_widgets.rb#L364-L368)), make the PR and then update the [yast2 dependency](https://github.com/yast/yast-registration/blob/9c04bc3d0c44488425f056f781d0c091a9efb10e/package/yast2-registration.spec#L43) here, right?

yes

jreidinger: @dgdavid regarding that widget, I find it very inconsistent with rest of other widgets regarding selecting/unselecting checkboxes. For other widgets and I mean mainly https://github.com/yast/yast-yast2/blob/b0eebf01c5485448393933563082cb908de0d3fc/library/cwm/src/lib/cwm/common_widgets.rb#L214 which is the closest to this checkbox list you can easily set list of selected items and also easily get info what is selected in that widget. See also similar table widget. So what I do not like is inconsistency with other widgets.

Also as I said consider you want to reuse your code aka widgets. What it means? you can use widgets which does not provide much value. You can reuse list container and can reuse details container.

I think that details widget can be used standalone without issues.
But can you reuse standlone CheckboxList? I think it does not provide much value. So if I want to reuse it I am force to use MasterDetailSelector, but what if I do not want details? Then I have to write it again from scratch and CheckboxList itself does not provide much.

Also consider what provides standalone CheckboxItem. Does it make sense to use it anywhere else beside in richtext that wants to display checkboxes? I think not, so it does not make much sense to not pass it into CheckboxList and benefit from it.

About that _representational widgets_ I worry it is not CWM philosophy ( and I sometimes are not much happy about it ). CWM is more like glue/controller, then just widget. It was always designed as controller, that just specify type of widget ( so also "creates" widget ), but its main goal is init/store methods that sets initial values and store what user specify at the end. And often also do validation.

Maybe it helps if someone outside look at it like @imobachgs or @ancorgs what is their opinion about it and what will fit the best. If we should try dumb widgets or keep consistency with rest of CWM.

dgdavid: > I think that details widget can be used standalone without issues.
But can you reuse standlone CheckboxList? I think it does not provide much value. So if I want to reuse it I am force to use MasterDetailSelector, but what if I do not want details? Then I have to write it again from scratch and CheckboxList itself does not provide much.

That's true. Rigth now, `CheckboxList` does not make sense without `MasterDetailSelector`.

> Also consider what provides standalone CheckboxItem. Does it make sense to use it anywhere else beside in richtext that wants to display checkboxes? I think not

Also true. It simple produces a checkbox _simulation_ in a format understandable by a RichText.

> About that representational widgets I worry it is not CWM philosophy ( and I sometimes are not much happy about it ). CWM is more like glue/controller, then just widget. It was always designed as controller, that just specify type of widget ( so also "creates" widget ), but its main goal is init/store methods that sets initial values and store what user specify at the end. And often also do validation.

Thanks for that explanation. I knew is not the CWM philosophy but now I have more details.

> Maybe it helps if someone outside look at it like @imobachgs or @ancorgs what is their opinion about it and what will fit the best. If we should try dumb widgets or keep consistency with rest of CWM.

All opinions are welcome, of course. But I guess it will not pay off only for this screen. Probably we will need more valid scenarios to justify starting to use a new approach. If so, I'll ask for help / pair-programming to fit CWM requirements.

imobachgs: I understand both points of view and I would like to start a discussion (not in this PR, of course) about the overall CWM philosophy. Maybe it is time to unify our vision about it.

In `yast2-storage-ng` you can see many widgets having a separate controller, in `yast2-configuration-management` we only use the `#store` method in a single widget, etc. I tend to think that "dump" widgets are easier to reuse... as widgets (not as controllers with all the associated logic). But having a discussion and setting some rules would help to move forward.

What do you think?

jreidinger: It is fine for me to unify view about usage especially if we agreed, we can add some helpers to make its usage easier.

joseivanlopez: I also agree that a discussion is needed. IMHO, we need to re-think our UI tools completely, from libyui to CWM. I don't say we have to kill everything, but maybe start looking to other approaches as a possible future improvement.

Regarding this PR, I agree with Josef about not breaking CWM philosophy. If we include something under CWM namespace, I would try to follow the same approach as the rest of existing widgets.

dgdavid: So, where/when do you think we should continue talking about that?

jreidinger: ideally try to organize meeting with all interested people, create notes from it and ideally result should do design document that will be then implemented. Something like we do team charter :)

joseivanlopez: Well, IIUC in the discussion during the retrospective meeting, the idea is a) to continue with the RichText approach for this new widget, and b) to arrange a specific meeting to discuss about how to improve libyui to avoid this pain.

For a) I guess that POs should decide when to continue with the implementation of this new CWM widget. And for b) we can try proposing a date/hour. But fist, I think we need to write some starting document (gist ?) with the problems discussed and maybe with some possible solution we have already mentioned. What do you think?

dgdavid: Yes, you remember correctly. But the problem I see is that this is a different topic. Related but different.

Here we were talking about 

> to re-think our UI tools completely, from libyui to CWM

and more specifically to move forward (or not) the idea of having _dumb_ widgets which only responsibility will be to present the data.

dgdavid: After our last retro meeting, I was talking with the PO to see if we could postpone this now that we [fixed the most important problem in `yast2-packager`](https://github.com/yast/yast-packager/pull/511) (the other candidate module to make use of this new set of dialog/widgets). But not, we need to move it forward since we want

* to unify those two dialogs
* to fix the other issues in those selectors

I still believe that a more _state less_ approach would be better to deal with YaST widgets, keeping the business logic outside the widgets, which only should take care of the visual data representation and user interaction.

However, I'd like to finish this endless task as soon as possible. So, I am going to ask you to help me move this forward. Ideally, a pair-programming session to adapt this PR (if possible) or create a new one is quite welcome.

We can continue the discussion later, as Iván pointed out in previous comments.

Thank you so much for your support, reviews, comments and suggestions folks.

dgdavid:
As said [above](#discussion_r388774481), we opted for a solution more aligned with the CWM philosophy.

So, closing this PR in favor of

* yast-yast2, CWM::RichText able to keep the scroll -- https://github.com/yast/yast-yast2/pull/1027
* yast-yast2, a new widget to manage multi status selections -- https://github.com/yast/yast-yast2/pull/1029
* yast-registartion, selection improved based in the previous work -- https://github.com/yast/yast-registration/pull/484
* yast-packager, selection improved based in the previous work -- https://github.com/yast/yast-packager/pull/515

PR #486, opened 2020-03-23
Suppress feedback

jreidinger:
trello: https://trello.com/c/AyGNkbK3/1708-3-sle15-sp2-p1-1165705-not-able-to-setup-network-when-connection-with-registration-server-failed-a-blocking-popupfeedback

bugzilla: https://bugzilla.suse.com/show_bug.cgi?id=1165705

Issue: Feedback is shown when registering system. When network is not configured it raise popup that allow to configure it. When clicked yes to configure it, popup that mentions "registering system" is still shown on top.

Fix: in yast2 was implemented in call to suppress feedback - https://github.com/yast/yast-yast2/pull/1034 . And this call is used here before calling network client.

### DIFF
--- a/package/yast2-registration.changes
+++ b/package/yast2-registration.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Mon Mar 23 14:34:05 UTC 2020 - Josef Reidinger <jreidinger@suse.com>
+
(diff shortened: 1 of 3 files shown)


jreidinger:
travis failed due to old image. spec dependencies ensure it works with rake osc:build

PR #488, opened 2020-03-26
[SP3] Fix misleading default_skipping_text (BSC#1167795)

cwickert:
https://bugzilla.suse.com/show_bug.cgi?id=1167795

For SLE 15 SP2, there is no 'Packages' media anymore and no need to
configure access to the manually.

Streamline text with online_skipping_text and make a small style fix
there.


lslezak:
As discussed in https://bugzilla.suse.com/show_bug.cgi?id=1167795 let's postpone for SP3.

jreidinger:
SP3 is here. So we need just adding changes.

lslezak (approved):

PR #494, opened 2020-05-20
Declining/refusal of an addon license means canceling all addons. (bsc#1169577)

mvidner:
- https://bugzilla.suse.com/show_bug.cgi?id=1169577
- https://trello.com/c/zRXvFHx3

Without this, an addon is still installed even if its license is declined.

In commit 1c7d041f6b3e71dce6a6f6062f10b1d6f0121754, PR https://github.com/yast/yast-registration/pull/440 ,
the idea was that we can unselect an addon whose license was declined, and
proceed with installing the remaining addons. But that does not work out with
autoselected addons ( https://github.com/yast/yast-registration/pull/335 https://github.com/yast/yast-registration/pull/338 ) where we take into account the product dependencies.

~~The new solution is to interpret a refusal of a license as the Abort button,
which ends installation of ALL selected addons.~~
Well that did not work out (because the dialog sequence has a dynamic ws_start) but now we have a better, simpler solution: declining a license is interpreted as the **Back** button, leading to the addon selection screen where you can unselect the offending one and still get the benefit of dependency resolution.

## Testing

Tested this manually:
- started with an unpatched SLE_15_SP1-x86_64-default.qcow2 image from Slenkins
- (in this state, declining a license immeidately aborts yast)
- `zypper up yast\*`
- `zypper in /tmp/yast2-registration-4.1.26-1.noarch.rpm`
- `yast scc` (first register)
- `yast scc select_extensions` (like `yast`, "Software > Add System Extensions or Modules")
- On x86_86 I could not reproduce the ppc64 case of an addon that has a licence without a regcode, it's one or the other. I used the Workstation Extensions (license AND regcode) and Package Hub (no license, no regcode)
- Without the fix, declining the WE license still asks for its regcode. With the fix, it goes back ald lets you unselect.

### DIFF
--- a/src/lib/registration/ui/addon_eula_dialog.rb
+++ b/src/lib/registration/ui/addon_eula_dialog.rb
@@ -48,6 +48,9 @@ module Registration
           false
         )
 
+        # Default: no EULA specified => accepted
+        eula_ret = :accepted
+
         addons.each do |addon|
           next unless addon.eula_acceptance_needed?
           next if addon.registered?
@@ -55,10 +58,13 @@ module Registration
           log.info "Addon '#{addon.name}' has an EULA at #{addon.eula_url}"
           eula_ret = accept_eula(addon)
 
-          return eula_ret if eula_ret != :next
+          # any declined license needs to be handled separately
+          break if eula_ret != :accepted
         end
 
-        :next
+        # go back if any EULA has not been accepted, let the user
+        # deselect the not accepted extension
+        eula_ret == :accepted ? :next : eula_ret
       end
 
     private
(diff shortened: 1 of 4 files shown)


mvidner:
Manual testing has shown that returning to the `"check"` wizard step is not good, since we should be returning to `workflow_start`, https://github.com/yast/yast-registration/blob/997324bc862107147bbe4dce40e102a1d63c0725/src/lib/registration/clients/inst_scc.rb#L332

That leads me to an even better solution: interpreting license refusal as the **Back** button. Will test.

src/lib/registration/ui/addon_eula_dialog.rb
```diff
@@ -48,17 +48,23 @@ def run
...
 
-          return eula_ret if eula_ret != :next
+          # any declined license needs to be handled separately
+          break if eula_ret != :accepted
         end
 
-        :next
+        # go back if any EULA has not been accepted, let the user
+        # deselect the not accepted extension
+        eula_ret == :accepted ? :next : eula_ret
```
imobachgs, line 67: :+1:

mvidner:
@imobachgs I have compared manual `COVERAGE=1 rake test:unit` runs between this branch and the base, SLE15-SP1, and there is no such drop. Coveralls is unreliable in this regard :(

PR #498, opened 2020-07-07
do not use directly profile

jreidinger:
trello: https://trello.com/c/kCfh7Ylo/1941-5-drop-references-to-profilecurrent

### DIFF
--- a/src/lib/registration/clients/scc_auto.rb
+++ b/src/lib/registration/clients/scc_auto.rb
@@ -89,6 +89,7 @@ module Registration
         # if there is no registration section like can happen during auto upgrade
         return unless settings
 
+        # Lazy load it as registration does not depend on ay, but scc_auto is run only in ay context
         Yast.import "AutoinstFunctions"
 
         # merge reg code if not defined in the profile but
@@ -198,6 +199,7 @@ module Registration
 
         products = Y2Packager::ProductControlProduct.products
 
+        # Lazy load it as registration does not depend on ay, but scc_auto is run only in ay context
         Yast.import "AutoinstFunctions"
 
         selected_product = Yast::AutoinstFunctions.selected_product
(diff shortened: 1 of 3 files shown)


src/lib/registration/clients/scc_auto.rb
```diff
@@ -198,7 +197,16 @@ def online_medium_config
         Yast::Packages.ImportGPGKeys
 
         products = Y2Packager::ProductControlProduct.products
-        ay_product = Yast::Profile.current.fetch("software", {}).fetch("products", []).first
+
+        Yast.import "AutoinstFunctions"
```
lslezak, line 201: Um, this one is part of the `autoyast2-installation` package which we do not `Require`. But as this is only called during (auto)installation then it is safe. But maybe it is still worth adding a comment so it's not accidentally copied elsewhere where it might not work.

jreidinger: yeah, it is already used in import. So I just copy it also there.

src/lib/registration/clients/scc_auto.rb
```diff
@@ -198,7 +197,16 @@ def online_medium_config
...
+
+        selected_product = Yast::AutoinstFunctions.selected_product
+        log.info "selected product #{selected_product.inspect}"
+        ay_product = if selected_product.respond_to?(:name)
```
imobachgs, line 205: So `Yast::AutoinstFunctions.selected_product` can return different things? According to the documentation, it is expected to return a [Y2Packager::Product](https://github.com/yast/yast-autoinstallation/blob/58f204a4ad7e47b59e892e3bb2fb9d4cb7d71f30/src/modules/AutoinstFunctions.rb#L84) instance. Perhaps I am missing something.

jreidinger: sadly it now can return both due to new medium layout where it can in some phases return controlProduct. I will adapt documentation. See https://github.com/yast/yast-autoinstallation/blob/58f204a4ad7e47b59e892e3bb2fb9d4cb7d71f30/src/modules/AutoinstFunctions.rb#L150

PR #499, opened 2020-07-09
Evaluating addon-list: Uses version without release for connecting SCC.

schubi2:
Due this change
https://github.com/yast/yast-registration/pull/454/files#diff-8abb084e7b29d0f192dcc6ab937bc948R140

The call "old" base_product_to_register returns "version_version" for the version whereas the "new" call installed_base_product returns "version_release".

While an upgrade SCC will be given a version_release number now which seems to be wrong.


schubi2:
First SLES15-SP2

PR #502, opened 2020-07-14
Clone SUSE registration settings

imobachgs:
Trello: https://trello.com/c/jzih5O43/1928-8-export-registered-add-ons-properly

## The problem

Until now, AutoYaST has been unable to export the registration settings. So if the user *clones* a registered system, the resulting control file will contain an `<add-on>` section similar to the example below:

```xml
<add_on_others config:type="list">
  <listentry>
    <alias>Basesystem_Module_15_SP2_x86_64:SLE-Module-Basesystem15-SP2-Updates</alias>
    <media_url><![CDATA[https://updates.suse.com/SUSE/Updates/SLE-Module-Basesystem/15-SP2/x86_64/update?MqjB_Y0Xzskf_Wg3etaBd607dRq5cWb7d4INl-Wh9GwKRKGKHLyt3HV1xq3Jre22bUlWAahwI...]]></media_url>
    <name>SLE-Module-Basesystem15-SP2-Updates</name>
    <priority config:type="integer">99</priority>
    <product_dir/>
  </listentry>
  <!-- many other add-ons -->
</add_on_others config:type="list">
```

Additionally, the profile will not contain the corresponding `<suse_register>` section. If the user tries to use such a profile, the result will be an error screen:

Screenshot_sle15sp2-2_2020-07-14_15:11:05

If the online medium is used, it will refuse to start the installation (as the registration information is missing).

## Proposed solution

This PR implements support to export the `<suse_register>` section, including the registered addons:

```xml
<suse_register>
  <addons config:type="list">
    <addon>
      <arch>x86_64</arch>
      <name>sle-module-basesystem</name>
      <release_type>nil</release_type>
      <version>15.2</version>
    </addon>
    <!-- the rest of addons -->
  </addons>
  <do_registration config:type="boolean">true</do_registration>
  <email>user@example.net</email>
  <install_updates config:type="boolean">false</install_updates>
  <reg_code>XXXXXXXXXXX</reg_code>
  <reg_server>https://scc.suse.com</reg_server>
  <reg_server_cert/>
  <slp_discovery config:type="boolean">false</slp_discovery>
</suse_register>
```

## TODO

- [x] Export registration settings.
- [x] Export the list of registered add-ons.
- [x] Export the `release_type` and `reg_code` in the list of add-ons.
- [x] Do not export the registered add-ons in the `<add_on_others>` section (https://github.com/yast/yast-add-on/pull/99)
- [ ] Try to determine `install_updates` if possible.
- [x] Filter out empty values, like `reg_server_cert`.


.rubocop.yml
```diff
@@ -2,6 +2,12 @@
 inherit_from:
   /usr/share/YaST2/data/devtools/data/rubocop_yast_style.yml
 
+AllCops:
+  TargetRubyVersion: 2.3
```
lslezak, line 6: You can switch to the newer Rubocop by changing the config file above to `rubocop-0.71.0_yast_style.yml`. That already uses `TargetRubyVersion: 2.6`...

lslezak:
A comment to the "Try to determine install_updates if possible." task: I think we cannot easily detect this from the installed system. Some customers install the system without updates (to have stable/reproducible installations) and install updates later. So having in stalled patches in the system does not help.

A similar problem exists with the `<reg_server_cert>` tag, in the installed system we do not know from where the SSL certificate was downloaded.

We will probably need to store these options at the end of the installation to some additional file...

imobachgs:
> A comment to the "Try to determine install_updates if possible." task: I think we cannot easily detect this from the installed system. Some customers install the system without updates (to have stable/reproducible installations) and install updates later. So having installed patches in the system does not help to figure this out.
> 
> A similar problem exists with the `<reg_server_cert>` tag, in the installed system we do not know from where the SSL certificate was downloaded.
> 
> We will probably need to store these options at the end of the installation to some additional file...

Thanks for the hints! I was not sure if extracting the values from the running system was possible at all. The idea of cloning a system is to export the *current* settings, not the settings we chose during installation. So maybe we could consider to simply export the default values for those elements.

.rubocop.yml
imobachgs: Yes, you are right. But I would need to do many changes and I rather prefer to not introduce more changes. But I will open a new PR later addressing it.

lslezak:
> The idea of cloning a system is to export the _current_ settings, not the settings we chose during installation.

I see. 

> So maybe we could consider to simply export the default values for those elements.

Then it makes sense to use some default values and document that. Esp. the problem with the SMT/RMT certificate needs to be documented.

PR #504, opened 2020-08-07
[WIP] [DEBUG] Parent fix

lslezak:
- This is just for testing, it seems that the Struct parent makes troubles in some cases, it might fail with `superclass mismatch for class RegCodes` error...


lslezak:
Closing, it turned out that the problem is elsewhere...

PR #508, opened 2020-10-01
Fixed upgrade with the `media_upgrade=1` option

lslezak:
- See https://bugzilla.suse.com/show_bug.cgi?id=1176563
- When using the `media_upgrade=1` boot option the base product repository was not initialized and there was no new product available to upgrade
- The fix is to call the `add_offline_base_product` method which is called when using the Offline medium also in the branch which handles the `media_upgrade=1` option.
- When using the Online medium we cannot add a product repository because there is no repository. That's the reason for a new error popup. (It's only displayed when using the Online medium with the `media_upgrade=1` option which is really rare so a missing translation should not be a problem.)

## Tests

- Tested manually, see the screenshots below

## Screenshots

The original error:

forced_media_upgrade_broken

Full medium with the fix (the upgrade continues):

forced_media_upgrade_fixed

Online medium with the fix (a new error is displayed):

forced_media_upgrade_online

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -12,7 +12,7 @@ Metrics/BlockNesting:
 
 # TODO: this need some non-trivial refactoring...
 Metrics/ClassLength:
-  Max: 470
+  Max: 480
 
 # TODO: this need some non-trivial refactoring...
 Metrics/CyclomaticComplexity:
--- a/src/lib/registration/ui/migration_repos_workflow.rb
+++ b/src/lib/registration/ui/migration_repos_workflow.rb
@@ -563,13 +563,24 @@ module Registration
       end
 
       # check the system status at upgrade and return the symbol for the next step
-      # @return [Symabol] workflow symbol, :skip => do not use the SCC/SMT/RMT upgrade
+      # @return [Symbol] workflow symbol, :skip => do not use the SCC/SMT/RMT upgrade
       #   (unregistered system or explicitly requested by user), :next =>
       #   continue with the SCC/SMT/RMT based upgrade
       def system_upgrade_check
         log.info "System upgrade mode detected"
         # media based upgrade requested by user
         if Yast::Linuxrc.InstallInf("MediaUpgrade") == "1"
+          # we cannot do medium based upgrade using the Online medium,
+          # it does not contain any packages
+          if Y2Packager::MediumType.online?
+            log.warn "The Online medium cannot be used with the media_upgrade=1 boot option"
+            Yast::Popup.LongMessage(online_media_upgrade)
+            return :abort
+          end
+
+          # add the Full medium base product repository
+          add_offline_base_product
+
           explicit_media_upgrade
           return :skip
         # the system is registered, continue with the SCC/SMT/RMT based upgrade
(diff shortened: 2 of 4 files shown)


.rubocop.yml
```diff
@@ -12,7 +12,7 @@ Metrics/BlockNesting:
 
 # TODO: this need some non-trivial refactoring...
 Metrics/ClassLength:
-  Max: 470
+  Max: 480
```
joseivanlopez, line 15: IMHO, this is a symptom of refactoring needed. Not a big deal yet, but I suggest to add a FIXME into *migration_repos_workflow.rb* to indicate the class is already too long and requires refactoring.

src/lib/registration/ui/migration_repos_workflow.rb
```diff
@@ -563,13 +563,24 @@ def update_releasever
...
       #   (unregistered system or explicitly requested by user), :next =>
       #   continue with the SCC/SMT/RMT based upgrade
       def system_upgrade_check
         log.info "System upgrade mode detected"
         # media based upgrade requested by user
         if Yast::Linuxrc.InstallInf("MediaUpgrade") == "1"
+          # we cannot do medium based upgrade using the Online medium,
+          # it does not contain any packages
+          if Y2Packager::MediumType.online?
+            log.warn "The Online medium cannot be used with the media_upgrade=1 boot option"
```
jreidinger, line 576: warn or error?

jreidinger (approved):
just np

src/lib/registration/ui/migration_repos_workflow.rb
```diff
@@ -753,6 +764,20 @@ def unregistered_online_message
...
 
+      # Informative message
+      # @return [String] translated message
+      def online_media_upgrade
+        # TRANSLATORS: Media upgrade forced but the Online medium was detected (1/2)
+        #   Message displayed during upgrade when the Online medium is used
+        #   with the "media_upgrade=1" boot option. The Online medium does not
+        #   contain any packages and it cannot be used in this case, Full medium
+        #   is required. Use the RichText format.
+        _("<h2>Online Medium</h2><p>The system cannot be upgraded using the Online " \
```
joseivanlopez, line 775: I am not sure whether the message is clear enough. I would indicate the source of the problem (i.e., because the media based upgrade was requested). I suggest something like:

"The media based upgrade was requested, but you are using the Online medium which does not provide any packages to install. Please use the Full medium instead of the Online."

What do you think?

jreidinger: suggestion sound fine for me, I just do not like that direct "you". Better impersonate with "but the Online medium is used"

src/lib/registration/ui/migration_repos_workflow.rb
lslezak: We should use warnings for not critical issues or user errors. There is an openQA test which scans the y2log for error after the installation.

src/lib/registration/ui/migration_repos_workflow.rb
joseivanlopez: What about "The media based upgrade was requested, but the Online medium is used. The Online medium does not provide any packages to install. Please use the Full medium instead of the Online."

joseivanlopez (approved):

PR #513, opened 2020-10-19
AY Addon: Taking the latest version if no version has been defined. T…

schubi2:
…aking system architecture if no architecture has been defined.

### DIFF
--- a/src/lib/registration/autoyast_addons.rb
+++ b/src/lib/registration/autoyast_addons.rb
@@ -21,6 +21,8 @@ module Registration
     attr_accessor :requested_addons, :selected_addons
 
     Yast.import "Report"
+    Yast.import "Pkg"
+    Yast.import "Arch"
 
     # Constructor
     # @param requested_addons [Array<Hash<String,String>>] the addons configuration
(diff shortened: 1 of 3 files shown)


src/lib/registration/autoyast_addons.rb
```diff
@@ -64,9 +66,18 @@ def register
     def select_addons(all_addons)
       # select the requested addons from the AY profile
       requested_addons.each do |addon|
-        server_addon = all_addons.find do |a|
-          a.identifier == addon["name"] && a.version == addon["version"] &&
-            a.arch == addon["arch"]
+        # Set architecture if it is not defined in the requested addon
+        requested_arch = addon["arch"] || Yast::Arch.architecture
```
lslezak, line 70: Unfortunately you cannot fallback to `Yast::Arch.architecture`, that does not work correctly in some cases. E.g. Arch can return `s390_64` while SCC expects `s390x`.

Use the [Y2Packager::ProductControlProduct::REG_ARCH](https://github.com/yast/yast-yast2/blob/master/library/packages/src/lib/y2packager/product_control_product.rb#L35) mapping like [this](https://github.com/yast/yast-yast2/blob/master/library/packages/src/lib/y2packager/product_control_product.rb#L61).

lslezak (commented):
Architecture mapping

src/lib/registration/autoyast_addons.rb
lslezak: So it could look like

```suggestion
        requested_arch = addon["arch"]
          || Y2Packager::ProductControlProduct::REG_ARCH[Yast::Arch.architecture]
          || Yast::Arch.architecture
```

And add `require "y2packager/product_control_product.rb"`.

lslezak (approved):

PR #517, opened 2020-10-27
Fix firstboot master

jreidinger:
merge of https://github.com/yast/yast-registration/pull/516

### DIFF
--- a/package/yast2-registration.changes
+++ b/package/yast2-registration.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Tue Oct 27 15:50:31 UTC 2020 - Josef Reidinger <jreidinger@suse.com>
+
+- Fix the latest change to work in firstboot when system is
+  registered (bsc#1178128)
+- 4.3.13
+
+-------------------------------------------------------------------
 Mon Oct 19 13:21:42 CEST 2020 - schubi@suse.de
 
 - AY Addon: Taking the latest version if no version has been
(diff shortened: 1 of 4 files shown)


mvidner (changes requested):
Please fix the failing test:

`rake osc:build` 

> ...
[   16s]   1) Yast::InstSccClient the system is already registered goes back to initial screen when aborting selection of url
[   16s]      Failure/Error: expect(subject.main).to eq(:next)
[   16s] 
[   16s]        expected: :next
[   16s]             got: :abort
[   16s] 
[   16s]        (compared using ==)
[   16s] 
[   16s]        Diff:
[   16s]        @@ -1 +1 @@
[   16s]        -:next
[   16s]        +:abort
[   16s]      # ./test/inst_scc_test.rb:63:in `block (3 levels) in <top (required)>'

mvidner:
BTW trying to build #516 fails at the start:

> unresolvable: nothing provides yast2-pkg-bindings >= 4.2.11 needed by yast2-update
      (got version 4.2.10-1.9)

jreidinger:
it is failing at https://github.com/yast/yast-registration/blob/master/test/inst_scc_test.rb#L63
reason is now when init is canceled, check registration returns abort. @lslezak any idea how this can be fixed?

lslezak:
I think we can remove whole test (https://github.com/yast/yast-registration/blob/master/test/inst_scc_test.rb#L53-L64), it seems completely invalid to me. Probably the workflow was a bit different in the past, currently the test looks weird.

It checks aborting the SLP URL selection but in a registered system. If the system is registered then the URL must have been already selected, you cannot abort the URL selection in an registered system. That does not make sense.

lslezak (approved):

mvidner (approved):
@jreidinger hint: it is better to just copy lslezak's comment to the commit message. it steals  code archeologists' jobs :)

PR #519, opened 2020-11-24
Theme fixes

jreidinger:
Draft of layout changes in registration dialog in installation. Several possible changes are shown. Basically each option represent evolution done by each commit.

## Vertical Spacing Only
The first commit is just remove of vertical spacing:
qt:
reg_qt
ncurses:
reg_ncurses

## Left Align
The second try to remove center align and instead align to left:
qt:
reg_qt_left

ncurses:
reg_ncurses_left

and minimal ncurses
reg_ncurses_left_min

## Left Align and Margin Box
The third try includes left align and margin box around
reg_qt_left_margin
reg_ncurses_left_margin

## Left Align and Horizontal Stretch
The forth try uses HStretch instead of Margin Box.
reg_qt_left_stretch
reg_ncurses_left_stretch
reg_qt_left_min_stretch

## Left Align and Horizontal Stretch and Network Button separate
It is same as previous one, just network button is moved to right edge. Looks like there is agreement that this layout is the best one from proposed.
reg_qt_left_stretch_nb
reg_ncurses_left_stretch_nb
reg_ncurses_left_stretch_min_nb

### DIFF
--- a/src/lib/registration/ui/base_system_registration_dialog.rb
+++ b/src/lib/registration/ui/base_system_registration_dialog.rb
@@ -179,7 +179,7 @@ module Registration
           network_button,
           VStretch(),
           product_details_widgets,
-          VSpacing(Yast::UI.TextMode ? 1 : 2),
+          VSpacing(Yast::UI.TextMode ? 1 : 0),
           registration_widgets,
           VStretch()
         )
@@ -187,7 +187,7 @@ module Registration
 
       def registration_widgets
         HSquash(
-          VBox(
+          HBox(
             RadioButtonGroup(
               Id(:action),
               VBox(
@@ -343,9 +343,7 @@ module Registration
 
         HSquash(
           VBox(
-            VSpacing(1),
             Left(Heading(SwMgmt.product_label(SwMgmt.find_base_product))),
-            VSpacing(1),
             label
           )
         )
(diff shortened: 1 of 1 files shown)


jreidinger:
@kwwii Can you check screenshots here what you prefer? On the second set I can also try to define instead min size for input fields and remove it horizontal stretching, so it can be shorter in qt and also ncurses.

jreidinger:
One more screenshot for default center aligned registration with high resolution which result in different view ( just part attached )

reg_qt_high_res

mvidner:
For testing of big-resolution layouts it may be helpful to scale down the qemu display so that it still fits your physical screen:

man qemu-kvm:

```
       During the graphical emulation, you can use special key combinations to
       change modes. The default key mappings are shown below, but if you use
       "-alt-grab" then the modifier is Ctrl-Alt-Shift (instead of Ctrl-Alt)
       and if you use "-ctrl-grab" then the modifier is the right Ctrl key
       (instead of Ctrl-Alt):

       Ctrl-Alt-f
           Toggle full screen

       Ctrl-Alt-+
           Enlarge the screen

       Ctrl-Alt--
           Shrink the screen

       Ctrl-Alt-u
           Restore the screen's un-scaled dimensions
```

lslezak:
After some discussion we decided to close it, this is just a low priority UI improvement...

PR #525, opened 2020-12-08
Updating warning popup if the user has skipped the registration.

schubi2:


src/lib/registration/ui/base_system_registration_dialog.rb
```diff
@@ -424,50 +424,19 @@ def default_skipping_headline
...
-
-        # these cannot be nil
-        if !media_name.empty? && !download_url.empty?
-          # TRANSLATORS: a popup message (2/3) the user wants to skip the registration
-          # %{media_name} is the media name (e.g. SLE-15-SP2-Full),
-          # %{download_url} is an URL link (e.g. https://download.suse.com)
-          warning += "\n\n" +
-            _("<p>For installation without registering the system use the\n"\
-              "%{media_name} medium from %{download_url}.</p>") %
-            { media_name: media_name, download_url: download_url }
```
lslezak, line 462: We are loosing the medium name and the URL in the message, I just [asked if that's OK](https://bugzilla.suse.com/show_bug.cgi?id=1177504#c9)....

lslezak (approved):
Let's merge it, if there are some more requests we can change the message later...

PR #527, opened 2021-01-19
Catch exceptions during cloning (bsc#1181075)

lslezak:
- https://bugzilla.suse.com/show_bug.cgi?id=1181075
- Display an error popup instead of crash
- The [catch_registration_errors](https://github.com/yast/yast-registration/blob/b021cdc4dd1891c25afc42bf87f99a4b28bb614a/src/lib/registration/connect_helpers.rb#L63) method catches exceptions and displays an error popup. It returns `true`/`false` depending on the success status of the block.
- 4.3.18


joseivanlopez (approved):
LGTM

PR #537, opened 2021-07-15
Show loading addons error

joseivanlopez:
## Problem

During the product migration, the registered but not installed addons are loaded in order to allow adding them. But sometimes the registration call fails for different reasons: timeout, internal error, etc. These exceptions are not controlled and they are raised to the user. 

* https://bugzilla.suse.com/show_bug.cgi?id=1187844
* (Internal) https://trello.com/c/epEQ3nt6/2579-3-publicsles-15-sp3-p3-1187844-build-maint-openqa-test-fails-in-upgradeselect

## Solution

As suggested by @lslezak, the *ConnectHelpers.catch_registration_errors* helper method is used. That helper displays an error popup and allows retrying in some cases, it can import the used SSL certificate, etc.

### DIFF
--- a/package/yast2-registration.changes
+++ b/package/yast2-registration.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Fri Jul 16 07:20:58 UTC 2021 - José Iván López González <jlopez@suse.com>
+
+- Show errors when loading addons fails (bsc#1187844).
+- 4.3.23
+
+-------------------------------------------------------------------
 Wed Apr 21 11:22:55 UTC 2021 - Imobach Gonzalez Sosa <igonzalezsosa@suse.com>
 
 - AutoYaST: do not crash when cloning a registered system with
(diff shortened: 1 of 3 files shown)


lslezak:
The idea in the whole module is that you in general do not care about the `SUSE::Connect` (or any other exceptions). At the places where you need to handle them you use the [`ConnectHelpers.catch_registration_errors`](https://github.com/yast/yast-registration/blob/032e600f331ded1a2673278aa0bf626c62731e9e/src/lib/registration/connect_helpers.rb#L63) helper method which displays an error popup and allows retrying in some cases.

See [this example](https://github.com/yast/yast-registration/blob/c5e4d3d1696320329f971c11a56d9c6eec487e7e/src/lib/registration/clients/inst_scc.rb#L170).

So I think the `migration_repos_workflow.rb` file should use this helper as well.

lslezak (approved):
:+1:

joseivanlopez:
https://build.suse.de/request/show/245249

PR #545, opened 2021-11-01
Fix missing product crash sp3

jreidinger:
related pr https://github.com/yast/yast-autoinstallation/pull/803
in general do not crash if no product is selected ( so selected is nil ).


lslezak (approved):
:+1:

PR #546, opened 2021-11-10
Add missing Yast::Logger includes

skazi0:
Some places implicitly used SUSE::Connect::Logger.


mvidner:
Thanks, Jacek!

Commenting on the 4 places:
- The widget inheriting from CWM is fine as it gets Yast::Logger from its base class
- SSL looks like an uncaught bug in exception handling, a classic one
- package search has unit tests which pass via SUSE::Connect::Logger which is the scary thing for me as I have no Idea how it gets there
- AY addons: haven't checked

skazi0:
I removed the CWM case (didn't have these sources at hand to check before).

lslezak (approved):
Thanks! :+1:

PR #548, opened 2021-11-22
Make tests compatible with libsuseconnect

skazi0:
Most functions return generic OpenStruct instead of specialized classes.
Some features (only used in tests) are missing (e.g. Product.to_triplet).
Asserts related to internal implementation details were removed or
converted to higher level (e.g. File mocks).


lslezak (approved):
Thank you for the update! :+1:

PR #556, opened 2021-12-21
Ruby3

jreidinger:


test/addon_selection_dialog_test.rb
```diff
@@ -113,6 +113,7 @@
     end
 
     it "works in textmode" do
+      pending "YaML loading of older OpenStruct is broken in ruby3" if RUBY_VERSION =~ /3\.0\.\d+/
```
lslezak, line 116: I'd specifically mention 3.0 as in 3.1 it should be fixed, also the RegExp should be fixed as that would also match 13.0 and others. :smile: 

```suggestion
      pending "YaML loading of older OpenStruct is broken in Ruby 3.0" if RUBY_VERSION =~ /^3\.0\.\d+/
```

lslezak (approved):

PR #557, opened 2021-12-31
Ruby3.1

jreidinger:
sadly that yaml load of older openstruct is still broken in ruby 3.1 . I will do deeper debugging after new year, but for now to unblock staging, lets keep it commented out.

### DIFF
--- a/package/yast2-registration.changes
+++ b/package/yast2-registration.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Fri Dec 31 16:11:06 UTC 2021 - Josef Reidinger <jreidinger@suse.com>
+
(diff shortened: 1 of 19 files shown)


teclator (approved):
LGTM

lslezak:
Um, as this effectively removes support for Ruby 3.0, shouldn't we add `Conflicts: ruby3.0` to the *.spec file? Just to be sure it is not accidentally used with Ruby 3.0?

Ruby 3.0 is still present in Factory... (https://build.opensuse.org/package/show/openSUSE:Factory/ruby3.0)

lslezak (approved):
Ok, it turned out that it actually works with all Rubies (2.5, 3.0 and 3.1)

PR #561, opened 2022-01-24
Use isbase

imobachgs:
Bug: [bsc#1194989](https://bugzilla.suse.com/show_bug.cgi?id=1194989)
Trello: https://trello.com/c/LX0JUO1E/

Starting in `SP4`, the `#base` method on objects coming from SCC does not seem to work anymore (it always returns `nil`, like any other non-existent method). Instead, using `isbase` works just fine.

See [comment 3](https://bugzilla.suse.com/show_bug.cgi?id=1194989#c3) in the bug report for further information.

### DIFF
--- a/package/yast2-registration.changes
+++ b/package/yast2-registration.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Mon Jan 24 15:29:29 UTC 2022 - Imobach Gonzalez Sosa <igonzalezsosa@suse.com>
+
+- Fix detection of base products coming from SCC (bsc#1194989).
+- 4.4.13
+
+-------------------------------------------------------------------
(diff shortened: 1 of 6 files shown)


package/yast2-registration.changes
```diff
@@ -1,3 +1,9 @@
+-------------------------------------------------------------------
+Mon Jan 24 15:29:29 UTC 2022 - Imobach Gonzalez Sosa <igonzalezsosa@suse.com>
+
+- Fix detection of base products coming from SCC (bsc#1194989).
```
jreidinger, line 4: I think there is somewhere documentation for communication with scc. So we probably need to update it and ask scc if it was intentional

lslezak (approved):
Thanks! :+1:

PR #564, opened 2022-01-31
[bsc#1194440] Do not try to read regcode when the short_name is unknown

imobachgs:
Bug report: [bsc#1194440](https://bugzilla.suse.com/1194440)

When running the auto-upgrade using the online medium, the product's short name is unknown when the registration configuration is being imported (there are no repositories yet), so it is not possible to read the reg codes from a USB stick. Ideally, we could include the short name in the control file. But, for now, we just avoid reading the reg codes in such a scenario.

### DIFF
--- a/src/lib/registration/clients/scc_auto.rb
+++ b/src/lib/registration/clients/scc_auto.rb
@@ -95,7 +95,7 @@ module Registration
         # merge reg code if not defined in the profile but
         # available from other sources
         product = Yast::AutoinstFunctions.selected_product
-        if product && !settings["reg_code"]
+        if product&.respond_to?(:short_name) && !settings["reg_code"]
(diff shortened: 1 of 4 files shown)


lslezak (approved):
Thanks for nice tests! :+1:

src/lib/registration/clients/scc_auto.rb
```diff
@@ -95,7 +95,7 @@ def import(settings)
         # merge reg code if not defined in the profile but
         # available from other sources
         product = Yast::AutoinstFunctions.selected_product
-        if product && !settings["reg_code"]
+        if product&.respond_to?(:short_name) && !settings["reg_code"]
```
kobliha, line 98: Nitpicking: this change might deserve a comment. It's not really understandable just from reading the code. Imagine someone trying to debug this in 2 years.

PR #569, opened 2022-04-12
Reload SUSEConnect certs in instsys (bsc#1195220)

skazi0:
## Problem

During installation (instsys context) the certificates are not installed
via SUSEConnect so explicit reloading of Golang's cert pool is needed.

- https://bugzilla.suse.com/show_bug.cgi?id=1195220
- https://openqa.suse.de/tests/8227446
- https://github.com/SUSE/connect-ng/pull/130 [required for this to work]

## Solution

Force reloading of SUSEConnect's cert pool after installing the certs.

## Testing

- *Tested manually*

PR #571, opened 2022-04-22
[SP4] Reload SSL certificates after import

lslezak:
## Problem

- The Go language needs explicitly reload the SSL cache after importing an SSL certificate.
- https://bugzilla.suse.com/show_bug.cgi?id=1195220
- See https://github.com/yast/yast-registration/pull/569

### DIFF
--- a/src/lib/registration/ssl_certificate.rb
+++ b/src/lib/registration/ssl_certificate.rb
@@ -103,6 +103,9 @@ module Registration
       # Cleanup
       FileUtils.rm_rf(TMP_CA_CERTS_DIR)
 
+      # Reload SUSEConnect internal cert pool (suseconnect-ng only)
+      SUSE::Connect::SSLCertificate.reload if defined?(SUSE::Connect::SSLCertificate.reload)
+
       # Check that last file was copied to return true or false
       File.exist?(new_files.last)
     end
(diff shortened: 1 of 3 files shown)


src/lib/registration/ssl_certificate.rb
```diff
@@ -103,6 +103,9 @@ def self.update_instsys_ca
       # Cleanup
       FileUtils.rm_rf(TMP_CA_CERTS_DIR)
 
+      # Reload SUSEConnect internal cert pool (suseconnect-ng only)
+      SUSE::Connect::SSLCertificate.reload if defined?(SUSE::Connect::SSLCertificate.reload)
```
jreidinger, line 107: well using defined for method is kind of strange, as usually is used `respond_to?` is there specific reason for it?

lslezak: AFAIK no...
@skazi0 ?

skazi0: I just couldn't decide which one to use :o)
Feel free to change to `respond_to?`

skazi0: I updated https://github.com/yast/yast-registration/pull/569 (master version)

jreidinger (approved):

PR #266, opened 2016-07-29
Fix certificates handling

imobachgs:
This PR should fix some certificates issues during installation:

* [bsc#981428](https://bugzilla.suse.com/show_bug.cgi?id=981428)
* [bsc#989787](https://bugzilla.suse.com/show_bug.cgi?id=989787)
* [bsc#991028](https://bugzilla.suse.com/show_bug.cgi?id=991028)

PR #574, opened 2022-05-04
[SP4] Fix instsys cert handling (bsc#1195220)

skazi0:
The new version of SUSEConnect doesn't support OpenSSL's "BEGIN TRUSTED
CERTIFICATE" format. Separate `extract` calls are needed to get certs
for "openssl" and "pem" directories.

## Problem

Failure to install custom SMT/RMT certificates during installation (with suseconnect-ng).

- [Bugzilla link](https://bugzilla.suse.com/show_bug.cgi?id=1195220)
- https://github.com/yast/yast-registration/pull/266

## Solution

Instead of extracting once and copying to "openssl" and "pem" directories, extract twice in respective formats.

## Testing

- *Tested manually*


src/lib/registration/ssl_certificate.rb
```diff
@@ -81,33 +78,17 @@ def self.download(url, insecure: false)
...
-        files.each do |file|
-          # FileUtils.cp does not seem to allow copying the links without dereferencing them.
-          Yast::Execute.locally("cp", "--no-dereference", "--preserve=links", file, subdir)
-          memo << File.join(subdir, File.basename(file))
-        end
-      end
-
-      # Cleanup
-      FileUtils.rm_rf(TMP_CA_CERTS_DIR)
+        "--overwrite", File.join(CA_CERTS_DIR, "openssl"))
```
lslezak, line 84: The goal of the code above is to keep the original content. The `/var/lib/ca-certificates` directory contains the initial trusted certificates. The `--overwrite` option causes removal of the previous content and losing the default SSL certificates.

That means we would have troubles accessing the other SSL secured sites.

Unfortunately after removing the `--overwrite` option the command fails with error `directory already exists`. So we need to extract the files into a temporary location and copy them to the correct target directory.

lslezak (commented):
Unfortunately this deletes the original `/var/lib/ca-certificates` files...

src/lib/registration/ssl_certificate.rb
skazi0: OK. Then we need the copy code. I'll add it back.

skazi0: done

lslezak (commented):
Please bump the version and add a `.changes` entry, otherwise LGTM.

lslezak (approved):
LGTM, thanks for the fix!

(I have tested it, and it works fine!)

PR #578, opened 2022-06-07
bsc#1200035: do no crash detecting activated products when credentials are missing

imobachgs:
## Problem

- [bsc#1200035](https://bugzilla.suse.com/show_bug.cgi?id=1200035)

When cloning a system, AutoYaST crashes if there are additional repositories (with no products) and the system is not registered.

## Solution

Check whether the system is registered before trying to search for activated products.

## Testing

- Added a new unit test
- Tested manually

### DIFF
--- a/package/yast2-registration.changes
+++ b/package/yast2-registration.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Tue Jun  7 12:37:04 UTC 2022 - Imobach Gonzalez Sosa <igonzalezsosa@suse.com>
+
+- Do not crash when cloning an unregistered system with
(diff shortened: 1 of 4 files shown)


mvidner:
NP: I'll just note that I don't see the connection between the triggering condition (" there are additional repositories (with no products) and the system is not registered.") and an exception called `MissingSccCredentialsFile`.

mvidner (approved):

lslezak (approved):

jreidinger (approved):

lslezak (approved):
Yes, this is a better solution! :+1:

PR #583, opened 2022-10-25
Improved logging, use "log.group" call (bsc#1204625)

lslezak:
## Problem

- Improve the registration log, group the registration steps
- Bugzilla: https://bugzilla.suse.com/show_bug.cgi?id=1204625

## Review Notes

- Use the "Hide whitespace" diff option otherwise the diff is too long and harder to read.
- Simply use [this link]()

## Solution

- Use the new `log.group` call from https://github.com/yast/yast-ruby-bindings/pull/287

## Testing

- Tested manually

### DIFF
--- a/src/lib/registration/clients/inst_scc.rb
+++ b/src/lib/registration/clients/inst_scc.rb
@@ -137,18 +137,28 @@ module Yast
 
     # run the addon selection dialog
     def select_addons
-      # FIXME: available_addons is called just to fill cache with popup
-      return :cancel if get_available_addons == :cancel
+      log.group "Select add-ons" do |group|
+        # FIXME: available_addons is called just to fill cache with popup
+        return :cancel if get_available_addons == :cancel
 
-      if Registration::Addon.find_all(@registration).empty?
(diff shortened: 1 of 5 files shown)


src/lib/registration/clients/inst_scc.rb
```diff
@@ -294,7 +308,11 @@ def registration_ui
     def workflow_aliases
       {
         # skip this when going back
-        "check"                  => ->() { registration_check },
+        "check"                  => lambda do
```
mvidner, line 311: Looking at this, and at the added `log.group`s in `pkg_manager` and `register_addons`... it seems obvious that we should tell `Sequencer.Run` to make a group for each *alias* it runs, right?

jreidinger (approved):

src/lib/registration/clients/inst_scc.rb
lslezak: Um, that would be nice, but unfortunately there is no human readable description of each step which I'd like to have in the log. We would need to extend the logic and the workflow data...

PR #584, opened 2022-10-31
Fix hash vs keyword arguments in RSpec expectations (bsc#1204871)

mvidner:
## Problem

- https://bugzilla.opensuse.org/show_bug.cgi?id=1204871 "- [Staging] rubygem-rspec-* 3.12 breaks various yast modules build"

New rspec-mocks distinguish between hash and keyword arguments passed to `with`.

The tested code does use hashes but we've used keywords in tests because
it's shorter to write. Fixed.

rspec-mocks 3.11.2 / 2022-10-25:

> Bug Fixes: Support keyword argument semantics when constraining argument
> expectations using with on Ruby 3.0+ with instance_double
> https://github.com/rspec/rspec-mocks/pull/1473

## Solution

Use hashes in tests instead of keywords

## Testing

- [x] Red-green tested with manually installed rspec 3.12.0

## Screenshots

Not affected

### DIFF
--- a/test/helpers_spec.rb
+++ b/test/helpers_spec.rb
@@ -166,14 +166,15 @@ describe Registration::Helpers do
   describe ".write_config" do
     it "writes the current configuration" do
       url = "https://example.com"
+      config = {
+        url:      url,
+        insecure: false
+      }
       expect(Registration::UrlHelpers).to receive(:registration_url) \
         .and_return(url)
       expect(Registration::Helpers).to receive(:insecure_registration) \
         .and_return(false)
-      expect(SUSE::Connect::YaST).to receive(:write_config).with(
-        url:      url,
-        insecure: false
-      )
+      expect(SUSE::Connect::YaST).to receive(:write_config).with(config)
 
       Registration::Helpers.write_config
     end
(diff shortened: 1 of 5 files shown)


test/helpers_spec.rb
```diff
@@ -166,14 +166,15 @@
...
+        insecure: false
+      }
       expect(Registration::UrlHelpers).to receive(:registration_url) \
         .and_return(url)
       expect(Registration::Helpers).to receive(:insecure_registration) \
         .and_return(false)
-      expect(SUSE::Connect::YaST).to receive(:write_config).with(
-        url:      url,
-        insecure: false
-      )
```
jreidinger, line 176: well, I am not sure if old form is wrong as it is one of ruby feature to auto convert hash if it is only argument. But maybe lost with ruby3?

jreidinger (approved):

test/helpers_spec.rb
mvidner: Since Ruby 3 hashes and keyword arguments are distinct: https://www.ruby-lang.org/en/news/2019/12/12/separation-of-positional-and-keyword-arguments-in-ruby-3-0/

PR #588, opened 2023-09-13
[SLE15-SP1] Update unit tests to work with the new suseconnect-ng

lslezak:
## Problem

- The backport support for the new suseconnect-ng did not work properly, the unit tests failed

## Solution

- This is mostly backport of these commits:
  - https://github.com/yast/yast-registration/commit/080ad0d5497adacfa98f8c845217dfa397365236
  - https://github.com/yast/yast-registration/commit/56e073eed6c391c38b7cb46a35dd5cbdc64eac28
  - https://github.com/yast/yast-registration/commit/a56ab06d6876f2fc097a0ff96bc32c29212600da

## Testing

- Tested manually using `osc build` in the [maintenance project](https://build.suse.de/package/show/SUSE:Maintenance:29622/yast2-registration.SUSE_SLE-15-SP1_Update) (_edit_: link corrected)

## Notes

- No version change, the package has not been released yet because of the build failures
- Just update the package and resubmit manually again

PR #589, opened 2023-09-13
[SLE15-SP2] Update unit tests to work with the new suseconnect-ng

lslezak:
## Problem

- The backport support for the new suseconnect-ng did not work properly, the unit tests failed

## Solution

- This is the same fix as in #588
- If you reviewed that then just check commit https://github.com/yast/yast-registration/commit/f85280eb0faa6473b27d7e213966ad38194a10c7, that's the only difference


mvidner (approved):
Thanks!

lslezak:
Submitted manually in https://build.suse.de/request/show/307465

PR #590, opened 2023-09-13
[SLE15-SP3] Update unit tests to work with the new suseconnect-ng

lslezak:
## Problem

- The backport support for the new suseconnect-ng did not work properly, the unit tests failed

## Solution

- This is the same fix as in #589 (and #588)
- If you reviewed them then just check commit https://github.com/yast/yast-registration/commit/97207bda594fb67045bcc5283d90d039348540a9 , that's the only difference

## Notes

- The unit tests in CI fail because it still uses the old suseconnect, I'll update the Docker image when it is accepted in SP3
- But building the package locally works fine


jreidinger (approved):

mvidner (approved):
Local build passed also for me, LGTM

lslezak:
Submitted manually in https://build.suse.de/request/show/307466

PR #591, opened 2023-11-22
Adapt to SCC API Change: 'base' -> 'isbase' [SLE-15-SP1]

shundhammer:
## Target Branch

_This is for **SLE-15-SP1**._ This is the **original PR** which will be merged forward to SLE-15-SP2 and SP3.

See section _Related PRs_ below for other target branches.

## Bugzilla

https://bugzilla.suse.com/show_bug.cgi?id=1217317

## Problem

Migration to a newer SLE-15 SP fails with an unrecoverable exception

_**Caught error: NoMethodError: "undefined method `friendly_name' for nil:NilClass"**_

Notice that this happens with the _old_ release while preparing the migration, even before the installation media with the newer release is booted.

## Cause

The API of SCC changed. The previous boolean field / method `base` was renamed to `isbase`.

This should never have happened in the first place during the lifetime of SLE-15. But now that the damage is done, we have to adapt to the change to minimize customer impact.

## Fix

Cherry-picked @imobachgs's fix 431d937b78c209c0d35ce911a98f9f05978e77ca which does a global `s/base/isbase/` in this source repo for the code and the test case.

## Affected Scenarios

@lslezak wrote on IRC:
> This code is fortunately used in offline migration only, so it could only happen with new QU media.

## Target Branches and Related PRs

- Original PR where this patch was cherry-picked from: https://github.com/yast/yast-registration/pull/561
- For SLE-15-SP2:  _TBD_
- For SLE-15-SP3:  _TBD_

- SLE-15-SP4, -SP5, -SP6 are not affected; the original patch was for SP4 and merged forward.

- SLE-15-GA went out of support.

### DIFF
--- a/package/yast2-registration.changes
+++ b/package/yast2-registration.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Wed Nov 22 09:19:08 UTC 2023 - Stefan Hundhammer <shundhammer@suse.com>
+
+- Adapted to SCC API change 'base' -> 'isbase' (bsc#1217317):
+  Cherry-picked igonzalezsosa's commit 431d937b78c209c0d35
+- 4.1.28
+
+-------------------------------------------------------------------
 Wed Aug  2 08:44:51 UTC 2023 - Ladislav Slezák <lslezak@suse.com>
 
 - Switch to the new SUSEConnect-ng (bsc#1212799)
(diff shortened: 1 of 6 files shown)


lslezak (approved):
LGTM, thanks! :+1:

shundhammer:
Submitted to IBS as https://build.suse.de/request/show/313363

